PR #5, opened 2021-06-04
[WIP] Akka.Coordination.Azure

### LINKED ISSUE #2: Lease: Akka.Storage.Blob Lease
**Is your feature request related to a problem? Please describe.**
We should add an [Akka.Coordination `Lease` implementation](https://getakka.net/api/Akka.Coordination.Lease.html) that uses a n [Azure Storage Blob Lease](https://docs.microsoft.com/en-us/rest/api/storageservices/lease-blob) as a coordination mechanism.

**Describe the solution you'd like**
A `Lease` that acquires an Azure Blob Lease and quickly releases it, for use inside [Akka.Cluster split brain resolvers](https://getakka.net/articles/clustering/split-brain-resolver.html) et al.

**Additional context**
This is going to require some pretty intensive integration testing, which can probably be done using [Azurite Storage Emulator](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azurite) inside a Docker container.

Aaronontheweb:
close #2


Aaronontheweb:
I think I can simplify the actor structure here quite a bit - but I do need to get some payload data in and out of the blob in order to know who the current owner of the lease is.

Arkatufus:
Superceeded by #865

PR #7, opened 2021-06-05
[WIP] Akka.Http.Core shim

ismaelhamed:
The biggest, and probably only, obstacle for porting the `Akka Management` (and the other packages that build on top of it) is the absence of the `Akka Http` module in Akka.NET. 

As you know, `Akka Http` leverages StreamIO underneath, but AFAIK you guys are still working on it as part of the `Artery` port. So a few months back I thought of the next best thing that I could come up with: creating a shim of `Akka Http` on top of Kestrel. 

Obviously this will lack backpressure and all, but it'd allow to port pretty much the whole `Akka Http` and have a compatible surface API while keeping the Kestrel part as an implementation detail inside the `Http` extension. This way, it'll be pretty straightforward to swap this shim with a proper implementation of the `Akka Http` in the future.

Currently, it has most of the Core Dsl (so no routes or directives) ported, and already gives the ability to create a very simple web server. The following  sample `Program.cs` should work right off the bat with the code in this branch:

```csharp
internal class Program
{
    private static readonly HttpResponse NotFound = HttpResponse.Create()
        .WithStatus(404)
        .WithEntity("Unknown resource!");
    
    public static void Main(string[] _)
    {
        var system = ActorSystem.Create("HelloAkkaHttpServer");
        
        // sync handler
        static HttpResponse RequestHandler(HttpRequest request) => request.Path switch
        {
            "/" => HttpResponse.Create().WithEntity(ContentTypes.TextHtmlUtf8, "<html><body>Hello world!</body></html>"),
            "/ping" => HttpResponse.Create().WithEntity("PONG!"),
            _ => NotFound
        };            

        var bindingTask = Http.Get(system).NewServerAt("localhost", 8085).BindSync(RequestHandler);
        bindingTask.WhenComplete((binding, exception) =>
        {
            if (binding != null)
            {
                var address = (DnsEndPoint) binding.LocalAddress;
                system.Log.Info("Server online at http://{0}:{1}/",
                    address.Host,
                    address.Port);

                // make sure Akka HTTP is shut down in a proper way
                binding.AddToCoordinatedShutdown(10.Seconds(), system);
            }
            else
            {
                system.Log.Error(exception, "Failed to bind HTTP endpoint, terminating system...");
                system.Terminate();
            }
        });

        Console.ReadLine();
    }
}
```

### DIFF
--- a/Akka.Management.sln
+++ b/Akka.Management.sln
@@ -18,6 +18,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Akka.Management", "src\mana
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Akka.Management.Tests", "src\management\Akka.Management.Tests\Akka.Management.Tests.csproj", "{5FD15B48-DECC-4742-AB9C-6F8517B12C6D}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Akka.Http.Shim", "src\management\Akka.Http.Shim\Akka.Http.Shim.csproj", "{DC9C497E-4B8D-4D32-A1D5-569E349BDE17}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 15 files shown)


Aaronontheweb:
@ismaelhamed does this need to be its own set of packages independent of Akka.Management?

Arkatufus:
Merged together in #13

PR #8, opened 2021-06-05
[WIP] Introduce Akka Management module

ismaelhamed:
This is most of the core `Akka Manangement` module (healthchecks are missing, though) already ported. I started working on it a few months back but had to put it aside. Since it looks like you guys are starting to work on it I thought you might want to reuse some of this.

Anything related to `Akka Http` (including `routes` and `directives`, as part of the High Dsl) is missing, so it won't compile. Hence PR #7 , which could be an option.

### DIFF
--- a/src/management/Akka.Management/Akka/Predef.cs
+++ b/src/management/Akka.Management/Akka/Predef.cs
@@ -1,0 +1,10 @@
+﻿namespace Akka
+{
+    public static class Predef
+    {
+        /// <summary>
+        /// Identity function to conform with the JVM API
+        /// </summary>
+        public static T Identity<T>(T x) => x;
+    }
+}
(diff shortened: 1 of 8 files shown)


src/management/Akka.Management/Akka/Predef.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Akka
+{
+    public static class Predef
+    {
+        /// <summary>
+        /// Identity function to conform with the JVM API
+        /// </summary>
+        public static T Identity<T>(T x) => x;
```
ismaelhamed, line 8: This should be part of `Akka` itself

Aaronontheweb: What does this do?

Arkatufus:
Merged together in #13

src/management/Akka.Management/Akka/Predef.cs
ismaelhamed: @Aaronontheweb this is just an [identity function](https://stackoverflow.com/questions/1797502/is-there-a-scala-identity-function)

PR #9, opened 2021-07-09
Introduce Akka.Discovery.AwsApi package

Arkatufus:


Arkatufus:
@Aaronontheweb I'm not sure if this is the correct way of doing this, can you give this a check?

Aaronontheweb (changes requested):
LGTM - can you please add a detailed `README.md` in this directory which explains how to use this plugin?

Arkatufus:
Will add the README.md and documentations, and most probably change the ECS discovery to internal so no one uses it

Aaronontheweb:
Looks good to me

Aaronontheweb (approved):

PR #13, opened 2021-07-30
Introduce Cluster.Bootstrap

Arkatufus:
This includes @ismaelhamed excellent Akka.Http shim (#7) and Akka.Management (#8) port, modified to work together without using any Akka.Streams


src/management/Akka.Http.Shim/Dsl/Http.cs
```diff
@@ -0,0 +1,135 @@
...
+        {
+            var localRoute = route;
+            var host = WebHost.CreateDefaultBuilder()
+                .SuppressStatusMessages(true)
+                .ConfigureServices(services =>
+                {
+                    services.AddLogging(builder => builder.AddFilter("Microsoft", LogLevel.Debug));
+                    services.AddRouting();
+                })
+                .Configure(app =>
```
Arkatufus, line 58: Here is a forward compatibility problem with the shim.
In the new ASP.NET, the `Configure(Action<IApplicationBuilder>)` method is burried and is not exposed as an API, the only way to access it is by implementing it inside a `Startup` class, which limits the way we can programatically assemble a server.

Arkatufus: Turns out that this is not a problem, the missing `Configure` method only applies to the new `IHostBuilder` API, it is still available with the older `IWebHostBuilder` API

ismaelhamed:
Love it that you have started working on this!

src/management/Akka.Http.Shim/Dsl/Model/HttpMessage.cs
```diff
@@ -0,0 +1,166 @@
...
+            new HttpRequest(request);
+
+        private HttpRequest(Microsoft.AspNetCore.Http.HttpRequest request)
+        {
+            _request = request;
+            var input = new byte[Convert.ToInt32(request.ContentLength)];
+            request.Body.Read(input, 0, input.Length);
+
+            Entity = new RequestEntity(request.ContentType, ByteString.FromBytes(input));
+        }
```
ismaelhamed, line 72: Do we really want this? I was trying to avoid leaking anything from AspNetCore outside the Http extesion itself. By leveraging the [Request Features](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/request-features) via `HttpRequestMiddleware`, we can do these mappings at the middleware level.

Is there any limitations to that approach?

Arkatufus: Not really, I was just trying to reduce the complexity of the shim. Except for the possibility that Microsoft changes the API (yet again) is there a danger of doing it this way? Is there really a need to completely decouple our code from ASP.NET?

Aaronontheweb: Referenced this in a new issue: https://github.com/akkadotnet/Akka.Management/issues/17

Still think we should merge this and push a new beta package so we can start testing out some the functionality, with the expectation that one way or another the Akka.HTTP APIs will evolve from here (i.e. they are not yet frozen)

Arkatufus:
@ismaelhamed Yes, I'll close #7 and #8

PR #25, opened 2021-08-09
Add documentations to Cluster.Bootstrap and Akka.Management

Arkatufus:


src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/ClusterBootstrapSettings.cs
```diff
@@ -17,7 +17,7 @@ public ContactPointDiscoverySettings(Config bootstrapConfig)
 
                 ServiceName = discoveryConfig.GetString("service-name");
                 if(string.IsNullOrEmpty(ServiceName) || ServiceName == "<service-name>")
-                    ServiceName = Environment.GetEnvironmentVariable("AKKA_CLUSTER_BOOTSTRAP_SERVICE_NAME");
+                    ServiceName = Environment.GetEnvironmentVariable("AKKA__CLUSTER__BOOTSTRAP__SERVICE_NAME");
```
Aaronontheweb, line 20: LGTM

src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/README.md
```diff
@@ -0,0 +1,280 @@
...
+```
+
+Ensure that `seed-nodes` is not present in configuration and that either autoloading through config or start() is called on every node.
+
+The following configuration is required, more details for each and additional configuration can be found in the reference configuration:
+
+- `akka.management.cluster.bootstrap.contact-point-discovery.service-name`: a unique name in the deployment environment for this 
+  cluster instance which is used to lookup peers in service discovery. If unset, it will be derived from the ActorSystem name.
+- `akka.management.cluster.bootstrap.contact-point-discovery.discovery-method`: the intended service discovery mechanism 
+  (from what choices Akka Discovery provides). If unset, falls back to the system-wide default from akka.discovery.method.
```
Aaronontheweb, line 42: Should link to relevant Akka.Discovery documentation here: https://getakka.net/articles/discovery/index.html - although that documentation is likely out of date and incomplete as well.

src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/README.md
```diff
@@ -0,0 +1,280 @@
...
+
+## How It Works
+
+- Each node exposes an HTTP endpoint `/bootstrap/seed-nodes`. This is provided by Akka.Management.Cluster.Bootstrap and 
+  exposed automatically by starting Akka.Management.
+  - `/bootstrap/seed-nodes` will query its internal cluster state and returns a list of members that are either in the up, weakly up,
```
Aaronontheweb, line 48: I think your list formatting is off here

src/discovery/Akka.Discovery.AwsApi/README.md
```diff
@@ -66,37 +66,53 @@ akka.discovery {
...
+on hard wired seeds configuration. Some HOCON configuration is needed to make discovery work with Cluster.Bootstrap:
+
+```
+akka.discovery.method = aws-api-ec2-tag-based
+akka.management.http.routes = {
+    cluster-bootstrap = "Akka.Management.Cluster.Bootstrap.ClusterBootstrapProvider, Akka.Management.Cluster.Bootstrap"
+}
+```
+
+You then start the cluster bootstrapping process by calling:
```
Aaronontheweb, line 80: We need an end to end example somewhere that shows:

- Creation of `ActorSystem`
- Bootstrapping + Discovery calls
- A piece of output that demonstrates that the cluster has started successfully. A `Cluster.RegisterOnMemberUp` callback would do nicely.

Think of this as a piece of code that a user should be able to copy and paste into their application that will "just work" provided that their configuration and environments are correct. Please flesh this out to expand the context of what's going on at application startup.

Aaronontheweb: Also, you need to show right here which ports need to be open on each VM in order for this to work. Don't force the user to go digging around through the other documentation pages if you can pack it into a quick blurb here.

src/management/Akka.Management/README.md
```diff
@@ -0,0 +1,144 @@
...
+are not designed to be and should never be opened to the public.
+
+## Stopping Akka Management
+In a dynamic environment you might stop instances of Akka Management, for example if you want to free up resources taken by 
+the HTTP server serving the Management routes.
+
+You can do so by calling Stop() on AkkaManagement. This method return a Task to inform when the server has been stopped.
+
+```
+val management = AkkaManagement(system)
```
Aaronontheweb, line 43: replace with C# syntax

Aaronontheweb (changes requested):
Some changes needed

Aaronontheweb (approved):
LGTM

src/management/Akka.Management/README.md
Aaronontheweb: This sample is still written in Scala

PR #29, opened 2021-08-12
Add .NET 5.0 support to Akka.Http.Shim

### LINKED ISSUE #23: Akka.Http.Shim uses Microsoft.AspNetCore 2.2.0, which is not compatible with .NET 5.0
As of right now, we're still aiming for .NET Standard 2.0 support, hence why we're using Microsoft.AspNetCore 2.2.0, but it seems like there is no update planned for this package in the future. Including Akka.Http.Shim in a project targetting .NET 5.0 will compile just fine but will fail to run as it will be missing some transient dependencies.

To support .NET 5.0 and up, we would need to do a dual deploy for both .NET Standard 2.0 and .NET 5.0, problem is that they changed the API in the library, so we would need to do conditional coding, and we will have to test the shim in both framework.

Arkatufus:
Closes #23

There are some build script changes, need to make sure that PR validations are correct

### DIFF
--- a/build-system/pr-validation.yaml
+++ b/build-system/pr-validation.yaml
@@ -27,7 +27,7 @@ jobs:
     parameters:
       name: 'linux_pr'
       displayName: 'Linux PR Validation'
-      vmImage: 'ubuntu-16.04'
+      vmImage: 'ubuntu-18.04'
       scriptFileName: ./build.sh
       scriptArgs: all
       outputDirectory: 'bin/nuget'
--- a/src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/Akka.Management.Cluster.Bootstrap.csproj
+++ b/src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/Akka.Management.Cluster.Bootstrap.csproj
@@ -1,7 +1,10 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFrameworks>$(LibraryFramework);net5.0</TargetFrameworks>
+        <TargetFrameworks>$(LibraryFramework);$(NetFramework)</TargetFrameworks>
+        <Description>Cluster bootstrap module for Akka.NET</Description>
+        <PackageTags>$(AkkaPackageTags)</PackageTags>
+        <GenerateDocumentationFile>true</GenerateDocumentationFile>
     </PropertyGroup>
 
     <ItemGroup>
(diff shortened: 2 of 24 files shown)


Arkatufus:
Adding .NET FX 4.7.1 causes the spec to become racy

build-system/pr-validation.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
     parameters:
       name: 'linux_pr'
       displayName: 'Linux PR Validation'
-      vmImage: 'ubuntu-16.04'
+      vmImage: 'ubuntu-18.04'
```
Aaronontheweb, line 30: LGTM

src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/Akka.Management.Cluster.Bootstrap.csproj
```diff
@@ -1,7 +1,10 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFramework>$(LibraryFramework)</TargetFramework>
+        <TargetFrameworks>$(LibraryFramework);$(NetFramework)</TargetFrameworks>
```
Aaronontheweb, line 4: Do we have any `#if`s or separate sets of dependencies for these packages?

src/management/Akka.Http.Shim/Akka.Http.Shim.csproj
```diff
@@ -16,7 +18,8 @@
 
   <ItemGroup>
     <PackageReference Include="Akka" Version="$(AkkaVersion)" />
-    <PackageReference Include="Microsoft.AspNetCore" Version="2.2.0" />
+    <PackageReference Condition="'$(TargetFramework)' == '$(LibraryFramework)'" Include="Microsoft.AspNetCore" Version="2.2.0" />
+    <FrameworkReference Condition="'$(TargetFramework)' == '$(NetFramework)'" Include="Microsoft.AspNetCore.App" />
```
Aaronontheweb, line 22: There we go - that answers my question from earlier.

src/management/Akka.Http.Shim/Dsl/Model/HttpMessage.cs
```diff
@@ -66,7 +66,11 @@ private HttpRequest(Microsoft.AspNetCore.Http.HttpRequest request)
         {
             _request = request;
             var input = new byte[Convert.ToInt32(request.ContentLength)];
+#if NET5_0
+            request.Body.ReadAsync(input, 0, input.Length).Wait();
```
Aaronontheweb, line 70: Ugly but this data should all be local to the machine by the time this call is made anyway.

Aaronontheweb (approved):
LGTM

src/management/Akka.Http.Shim/Dsl/Model/HttpMessage.cs
Arkatufus: Yes, there's no sync version for the new ASP.NET, everything are async

Aaronontheweb: I figured as much

PR #79, opened 2021-09-02
Fix erroneous failing probes after cluster formed.

Arkatufus:
Fixes:
- Stop all timers when cluster bootstrap coordinator and probe actors are stopped.
- Ignore all probe reports after probe are stopped.
- Add spec to check that cluster bootstrap coordinator are stopped after cluster have formed.
- Add a simple stress test console program to test rapid cluster forming and scale down.

Initial observation:
- There is a limit on how many nodes can be forced to spin up in rapid succession. 
  - A 60 node cluster is still achieveable with some remoting heartbeat failures.
  - A 75 node cluster fails to form because of networking congestion causing a lot of association errors.
- A cluster with nodes larger than 20 that are scaled down rapidly might spam a lot of `AssociationError` and `InvalidAssociationException` warnings after it is scaled down, even when the cluster is scaled using graceful shutdown.
- Numbers may vary from one computer to another, tested using a Ryzen 9 3900X machine with 32 GB of memory.
- This will not represent behaviour in real world production environment since all of the actor systems are spun inside a single machine and process.

### DIFF
--- a/src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/Internal/BootstrapCoordinator.cs
+++ b/src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/Internal/BootstrapCoordinator.cs
@@ -206,6 +206,16 @@ namespace Akka.Management.Cluster.Bootstrap.Internal
             StartPeriodicDecisionTimer();
         }
 
+        protected override void PostStop()
+        {
+            Timers.CancelAll();
+            foreach (var child in Context.GetChildren())
+            {
+                child.Tell(PoisonPill.Instance);
+            }
+            base.PostStop();
+        }
+
         private Receive Receive()
         {
             return message =>
@@ -228,7 +238,7 @@ namespace Akka.Management.Cluster.Bootstrap.Internal
             };
         }
 
-        private Receive Bootstrapping(IActorRef replyTo, string selfContactPointScheme)
+        protected virtual Receive Bootstrapping(IActorRef replyTo, string selfContactPointScheme)
         {
             return message =>
             {
@@ -431,8 +441,9 @@ namespace Akka.Management.Cluster.Bootstrap.Internal
             }
 
             var child = Context.Child(childActorName);
-            if (!child.Equals(ActorRefs.Nobody))
+            if (!child.IsNobody())
                 return child;
+            
             var props = HttpContactPointBootstrap.Props(_settings, contactPoint, baseUri);
             return Context.ActorOf(props, childActorName);
         }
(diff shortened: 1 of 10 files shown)


src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/Internal/BootstrapCoordinator.cs
```diff
@@ -206,6 +206,16 @@ protected override void PreStart()
             StartPeriodicDecisionTimer();
         }
 
+        protected override void PostStop()
+        {
+            Timers.CancelAll();
+            foreach (var child in Context.GetChildren())
+            {
+                child.Tell(PoisonPill.Instance);
```
Arkatufus, line 214: Make sure that all probe actors are told to shut down when the coordinator is shutting down

src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/Internal/HttpContactPointBootstrap.cs
```diff
@@ -94,47 +95,51 @@ public HttpContactPointBootstrap(ClusterBootstrapSettings settings, ResolvedTarg
                 var getTask = _http.GetAsync(_probeRequest);
                 getTask.ContinueWith(task =>
                 {
-                    if (task.IsCompleted)
```
Arkatufus, line 97: No code change, rearranged so that failures are tested first

src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/Internal/BootstrapCoordinator.cs
```diff
@@ -206,6 +206,16 @@ protected override void PreStart()
             StartPeriodicDecisionTimer();
         }
 
+        protected override void PostStop()
+        {
+            Timers.CancelAll();
```
Arkatufus, line 211: Cancel all running timers when actors are shut down.

src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/Internal/HttpContactPointBootstrap.cs
```diff
@@ -169,6 +177,13 @@ protected override void PreStart()
             Self.Tell(ProbeTick.Instance);
         }
 
+        protected override void PostStop()
+        {
+            _stopped = true;
+            Timers.CancelAll();
```
Arkatufus, line 183: The same in probe actors, cancel all timers when it is shut down.

Aaronontheweb (approved):
LGTM

PR #156, opened 2021-10-01
Management should not use IPAddress.Any as its default hostname

Arkatufus:
Change the way Akka.Management define default hostname.
- If not overriden or empty, check `akka.remote.dot-netty.tcp.public-hostname`, reason being that this is the settings being overriden by Akka.DockerBootstrap
- If `akka.remote.dot-netty.tcp.public-hostname` is empty, we use `Dns.GetHostName` instead.


src/management/Akka.Management/Resources/reference.conf
```diff
@@ -9,12 +9,13 @@ akka.management {
   http {
     # The hostname where the HTTP Server for Http Cluster Management will be started.
     # This defines the interface to use.
-    # InetAddress.getLocalHost.getHostAddress is used not overriden or empty
+    # akka.remote.dot-netty.tcp.public-hostname is used if not overriden or empty.
+    # if akka.remote.dot-netty.tcp.public-hostname is empty, Dns.GetHostName is used.
```
Aaronontheweb, line 13: Ah, so the `http.hostname` here would be advertised as the "public" hostname for connecting to this service?

Arkatufus: There's `http.hostname` and `http.bind-hostname`, the effective bind host name is `http.bind-hostname`, if it is empty then it falls back to `http.hostname`.

Arkatufus: This `http.hostname` has to match exactly to the host names resolved by discovery because it is used to determine if a resolved host name is "self" and used as a marker and check to determine if a node should join itself to start a cluster

Aaronontheweb: Ahhh got it - yes, you can't ever use `0.0.0.0` or the IPV6 equivalent for a public hostname. Will never work. Makes sense.

PR #231, opened 2021-10-28
Make Akka.Discovery.KubernetesApi Lookup method more robust

Arkatufus:
- Cache k8s client instead of re-instantiating it on every Lookup call
- Catch exceptions and transforms them into sensible warning logs and exception messages


Aaronontheweb (approved):
LGTM

PR #666, opened 2022-07-08
Add backing Azure Table data model and client

Arkatufus:
- Add proto file
- Add cluster member Azure data model
- Add Azure table client
- Modify AzDO build script to install Azurite locally

### DIFF
--- a/build-system/azure-pipeline.template.yaml
+++ b/build-system/azure-pipeline.template.yaml
@@ -26,6 +26,12 @@ jobs:
         inputs:
           version: 3.1.x
       # Linux or macOS
+      - bash: |
+          sudo npm install -g azurite@3.17.1
+          sudo mkdir azurite
+          sudo azurite --silent --location azurite --debug azirute/debug.log &
+        displayName: 'Install and Run Azurite (Linux)'
+        condition: in( variables['Agent.OS'], 'Linux', 'Darwin' )
       - task: Bash@3 
         displayName: Linux / OSX Build
         inputs:
@@ -34,6 +40,12 @@ jobs:
         continueOnError: true
         condition: in( variables['Agent.OS'], 'Linux', 'Darwin' )
       # Windows
+      - powershell: |
+          npm install -g azurite@3.17.1
+          mkdir azurite
+          Start-Process azurite.cmd -PassThru
+        displayName: 'Install and Run Azurite (Windows)'
+        condition: eq( variables['Agent.OS'], 'Windows_NT' )
       - task: BatchScript@1
         displayName: Windows Build
         inputs:
--- a/src/discovery/azure/Akka.Discovery.Azure/Akka.Discovery.Azure.csproj
+++ b/src/discovery/azure/Akka.Discovery.Azure/Akka.Discovery.Azure.csproj
@@ -1,0 +1,30 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>netstandard2.0</TargetFramework>
+        <LangVersion>8.0</LangVersion>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <None Remove="protobuf\ClusterMemberProto.proto" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <PackageReference Include="Akka.Cluster" Version="$(AkkaVersion)" />
+      <PackageReference Include="Azure.Data.Tables" Version="12.6.0" />
+      <PackageReference Include="Google.Protobuf" Version="3.21.2" />
+      <PackageReference Include="Grpc.Tools" Version="2.40.0">
+        <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+        <PrivateAssets>all</PrivateAssets>
+      </PackageReference>
+    </ItemGroup>
+
+    <ItemGroup>
+      <ProjectReference Include="..\..\..\management\Akka.Management\Akka.Management.csproj" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <Protobuf Include="protobuf\ClusterMemberProto.proto" Access="Internal" />
+    </ItemGroup>
+
+</Project>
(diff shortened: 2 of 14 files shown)


src/discovery/azure/Akka.Discovery.Azure/Model/ClusterMember.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        internal TableEntity Raw { get; }
+        internal ClusterMemberProto Proto { get; }
+        
+        public string ServiceName => PartitionKey;
+        public string Host => Proto.Host;
+        public IPAddress Address => IPAddress.Parse(Proto.Address);
+        public int Port => Proto.Port;
+        public DateTime Created => Proto.Created.ToDateTime();
+        public DateTime LastUpdate { get; }
```
Arkatufus, line 49: As per changes in the spec, the cluster member table data model is stripped to the bare minimum. The only fields exposed are the ones needed by the query for TTL scheme support.
All the rest of the supporting data are packed inside a byte array using protobuf instead, for data versioning reasons

build-system/azure-pipeline.template.yaml
```diff
@@ -26,6 +26,12 @@ jobs:
         inputs:
           version: 3.1.x
       # Linux or macOS
+      - bash: |
+          sudo npm install -g azurite@3.17.1
+          sudo mkdir azurite
+          sudo azurite --silent --location azurite --debug azirute/debug.log &
```
Aaronontheweb, line 32: You wanted to do this over `Docker.NET`?

src/discovery/azure/Akka.Discovery.Azure/Model/ClusterMember.cs
```diff
@@ -0,0 +1,130 @@
...
+        public static TableEntity CreateEntity(
+            string serviceName,
+            IPAddress address,
+            int port)
+        {
+            var now = DateTime.UtcNow;
+            var proto = new ClusterMemberProto
+            {
+                Host = Dns.GetHostName(),
+                Address = address.MapToIPv4().ToString(),
```
Aaronontheweb, line 70: Why have host and address as separate fields? Can't a node have either one or the other?

src/discovery/azure/Akka.Discovery.Azure/ClusterMemberTableClient.cs
```diff
@@ -0,0 +1,204 @@
...
+using Akka.Annotations;
+using Akka.Discovery.Azure.Model;
+using Akka.Event;
+using Azure;
+using Azure.Data.Tables;
+
+namespace Akka.Discovery.Azure
+{
+    [InternalApi]
+    public class ClusterMemberTableClient
```
Aaronontheweb, line 22: Does this need to be `public` or can we make it `internal`?

src/discovery/azure/Akka.Discovery.Azure/ClusterMemberTableClient.cs
```diff
@@ -0,0 +1,204 @@
...
+            return list.ToImmutableList();
+        }
+
+        internal async Task<bool> UpdateAsync(CancellationToken token = default)
+        {
+            if (!await EnsureInitializedAsync(token))
+                return false;
+
+            var original = _entity.LastUpdate;
+            _entity = _entity.Update();
```
Aaronontheweb, line 101: This updates the timestamp?

build-system/azure-pipeline.template.yaml
Arkatufus: No, the Azurite Docker image rarely got updated and it is linux only. Its the same reason why we switched to this way of doing it in Akka.Persistence.Azure.

Aaronontheweb: Got it, thanks

src/discovery/azure/Akka.Discovery.Azure/Model/ClusterMember.cs
Aaronontheweb: Ok, got it.

src/discovery/azure/Akka.Discovery.Azure/ClusterMemberTableClient.cs
Arkatufus: Yes, the ClusterMember class is immutable, this creates a new ClusterMember instance with an updated LastUpdate field.

src/discovery/azure/Akka.Discovery.Azure/ClusterMemberTableClient.cs
Aaronontheweb: Don't they all get started internally by the `Akka.Discovery` plugin though? i.e. the user doesn't have to launch any of those components.

src/discovery/azure/Akka.Discovery.Azure/Akka.Discovery.Azure.csproj
```diff
@@ -0,0 +1,30 @@
...
+
+    <ItemGroup>
+      <None Remove="protobuf\ClusterMemberProto.proto" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <PackageReference Include="Akka.Cluster" Version="$(AkkaVersion)" />
+      <PackageReference Include="Azure.Data.Tables" Version="12.6.0" />
+      <PackageReference Include="Google.Protobuf" Version="3.21.2" />
+      <PackageReference Include="Grpc.Tools" Version="2.40.0">
```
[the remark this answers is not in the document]
Aaronontheweb: LGTM then

src/discovery/azure/Akka.Discovery.Azure/ClusterMemberTableClient.cs
Arkatufus: Right, I'll change it.

Arkatufus: Done

Aaronontheweb:
@Arkatufus you might need to rebase on `dev` once https://github.com/akkadotnet/Akka.Management/pull/667 in order to get CI to run for this, FYI.

Arkatufus:
Done

Aaronontheweb:
@Arkatufus nice work - merged.

PR #706, opened 2022-07-28
Add guardian actor restart with backoff retry for failing DNS resolution and failing Az table client

Arkatufus:
Add a start-up retry delay when the guardian actor failed to initialize


Aaronontheweb (approved):
LGTM

PR #723, opened 2022-08-02
Fixes for Discovery.Azure

Arkatufus:
## Changes
- Update the OData query, missing "L" qualifier to mark that the numbers are to be read as an Int64, Azure would default to reading any numbers as an Int32 and throws an overflow parse error.
- Fix discovery `Lookup` method, it should not throw when the `Ask` operation timed out.


Aaronontheweb (approved):
LGTM

PR #728, opened 2022-08-04
Replace Dns.GetHostentry, it might not return the original IP address

Arkatufus:


src/discovery/azure/Akka.Discovery.Azure/Actors/AzureDiscoveryGuardian.cs
```diff
@@ -123,14 +124,19 @@ private bool Initializing(object message)
...
-                        _host = entry.HostName;
-                        _address = entry.AddressList
-                            .First(i =>
-                                !Equals(i, IPAddress.Any) &&
-                                !Equals(i, IPAddress.Loopback) &&
-                                !Equals(i, IPAddress.IPv6Any) &&
-                                !Equals(i, IPAddress.IPv6Loopback));
+                        if (IPAddress.TryParse(_settings.HostName, out _address))
+                        {
+                            if (_address.Equals(IPAddress.Any) || _address.Equals(IPAddress.IPv6Any))
```
Aaronontheweb, line 129: LGTM

src/discovery/azure/Akka.Discovery.Azure/Actors/AzureDiscoveryGuardian.cs
```diff
@@ -123,14 +124,19 @@ private bool Initializing(object message)
...
+                                throw new ConfigurationException($"IPAddress.Any or IPAddress.IPv6Any cannot be used as host address. Was: {_settings.HostName}");
+
+                            _host = Dns.GetHostName();
+                        }
+                        else
+                        {
+                            _host = _settings.HostName;
+                            _address = Dns.GetHostAddresses(_host)
```
Aaronontheweb, line 137: We need to add a test for when:

1. Hostname is valid for routing (externally)
2. Hostname is not assigned to anything locally via DHCP

This is an extremely common scenario in cloud environments outside of K8s since much of the network address translation happens at the software level above the VM. It's why we have the `public-hostname` and `public-port` settings in Akka.Remote.

__I would not worry about this right now though__ - save it for integration testing with Azure App Service.

Aaronontheweb (approved):
LGTM

PR #729, opened 2022-08-05
Replace Kestrel with Ceen

Arkatufus:
Replace Akka.HTTP.Shim internal HTTP server from Kestrel to Ceen.Httpd


Aaronontheweb:
Can you help spell out the benefits / reasoning for the change? I know we've talked about some of the HTTP overlap problems with Kestrel but still want to understand the decisions made on the PR.

Arkatufus:
This is a stop-gap temporary solution before we move back to Kestrel/IIS. Implementing this change is much faster in the short run, though we will need to revisit Kestrel/IIS integration in the future.

Aaronontheweb:
@dependabot rebase

PR #730, opened 2022-08-05
Add Setup classes to ClusterBootstrap

Arkatufus:


Arkatufus:
Should we make these Setup classes immutable?

Aaronontheweb:
> Should we make these Setup classes immutable?

Yes

Aaronontheweb:
@Arkatufus nevermind, keep the `Setup` classes mutable for the time being. We can refactor later on to use `record`s once we drop .NET Standard 2.0 support for Akka.NET 2.0.

PR #731, opened 2022-08-05
Add AkkaManagement Setup class

Arkatufus:


Arkatufus:
Should we make these Setup classes immutable?

src/management/Akka.Management/AkkaManagementSetup.cs
```diff
@@ -0,0 +1,100 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using Akka.Actor.Setup;
+
+namespace Akka.Management
+{
+    public sealed class AkkaManagementSetup: Setup
```
Aaronontheweb, line 16: Yes, we should make this immutable

src/management/Akka.Management/AkkaManagementSetup.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// The value will need to be from 0 to 65535.
+        /// </summary>
+        public int? Port { get; set; }
+        
+        /// <summary>
+        /// Use this setting to bind a network interface to a different hostname or ip
+        /// than the HTTP Server for Http Cluster Management.
+        /// Use "0.0.0.0" to bind to all interfaces.
+        /// </summary>
+        public string EffectiveBindHostname { get; set; }
```
Aaronontheweb, line 47: Want to stick with "public port" and "public hostname" conventions that we have established in Akka.Remote?

Aaronontheweb (changes requested):
Minor changes needed

src/management/Akka.Management/AkkaManagementSetup.cs
Aaronontheweb: Disregard - too much work for now.

Aaronontheweb (approved):

PR #742, opened 2022-08-09
[Discovery.Azure] Cleanup self member entry during ActorSystem shutdown

Arkatufus:
Add cleanup code during system shutdown, this lessen the amount of junk accumulating inside the Azure ClusterMember table, making cluster restart and scaling faster.


Aaronontheweb (approved):
Great idea. LGTM.

Issue #738, opened 2022-08-08
Akka.Hosting support for Akka.Management


Aaronontheweb:
**Is your feature request related to a problem? Please describe.**

Once the `Setup` classes et al are polished we should add Akka.Hosting support for:

- [x] Akka.Management (#747)
- [x] Akka.Management.Cluster.Bootstrap (#747)
- Coordination:
  - [x] Akka.Coordination.KubernetesApi (#784)
- Discovery:
  - [x] Akka.Discovery.AwsApi
  - [x] Akka.Discovery.Azure (#747)
  - [x] Akka.Discovery.KubernetesApi

**Describe the solution you'd like**

A composable API system akin to what we've built for Akka.Persistence.

**Describe alternatives you've considered**
N/A

**Additional context**

Do we want to cram all plugins into the same NuGet package? Have Akka.Discovery plugins use their own packages (i.e. same thing we've done for Akka.Persistence?)

SeanFarrow:
Separate NuGet packages would make sense. I'd be happy to help look at this.

Aaronontheweb:
@SeanFarrow we would welcome your help! @Arkatufus is going to spec something out here and stub out the NuGet package - from there we'll need help building the actual APIs themselves.

Arkatufus:
Added a starting PR for Akka.Hosting at https://github.com/akkadotnet/Akka.Management/pull/747

SeanFarrow:
Brilliant, what help do you need from here?

Arkatufus:
@SeanFarrow 

I've added a checklist/tasks of things that needed to be added in the issue description above, we'll check them as complete when the related PRs have been merged.
So far, we still need to add `Coordination.KubernetesApi`, `Discovery.AwsApi`, and `Discovery.KubernetesApi` support.

Thanks for the help :)

SeanFarrow:
OK, is it best to start with discovery.KubernetesApi, or Coordination.KubernetesApi?
What’s the difference between the discovery and coordination sides of this?
No worries, I love to help where I can and have wanted to for a while.

Thanks,
Sean.

Arkatufus:
Discovery is probably the easier one. 
Discovery helps with automated cluster formation, while Coordination handles managed leases for things like shared resource locking in a cloud/distributed system. 
For example, an actor in node A can claim exclusive use of a cloud resource by claiming a lease, other nodes in the cluster would not be able to access the same resource until node A releases the claimed lease.

[...]

PR #747, opened 2022-08-10
Add Akka.Hosting support

Arkatufus:
Part of #738

## Changes
Added Akka.Hosting support for:
- Akka.Management
- Akka.Management.Cluster.Bootstrap
- Akka.Discovery.Azure


SeanFarrow:
This looks good.

Once this is approved/merged, I'll happily work on adding extension methods and setup classes for the rest of the plugins (k8s etc).

SeanFarrow:
Personally, I would split it as then the *.hosting packages pull in only what they require.
As well as this, I can install a *.hosting package and get the required dependent packages without having to think about any dependencies.
Finally, if a direct or transient dependency is updated and a single hosting package is used, all consumers will have to update the *.hosting package (whether they use the particular system of discovery/coordination or not).
Thanks,
Sean.

src/hosting/Akka.Management.Hosting/Akka.Management.Hosting.csproj
```diff
@@ -0,0 +1,21 @@
...
+
+    <ItemGroup>
+      <ProjectReference Include="..\..\cluster.bootstrap\Akka.Management.Cluster.Bootstrap\Akka.Management.Cluster.Bootstrap.csproj" />
+      <ProjectReference Include="..\..\discovery\azure\Akka.Discovery.Azure\Akka.Discovery.Azure.csproj" />
+      <ProjectReference Include="..\..\management\Akka.Management\Akka.Management.csproj" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <PackageReference Include="Akka.Cluster.Hosting" Version="0.4.1" />
+      <PackageReference Include="Akka.Persistence.Azure.Hosting" Version="0.9.0" />
```
Aaronontheweb, line 18: Is this dependency necessary?

src/hosting/Akka.Management.Hosting/AzureDiscoveryExtensions.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// <param name="builder">
+        ///     The builder instance being configured.
+        /// </param>
+        /// <param name="setup">
+        ///     The <see cref="AzureDiscoverySetup"/> instance used to configure Akka.Discovery.Azure.
+        /// </param>
+        /// <returns>
+        ///     The same <see cref="AkkaConfigurationBuilder"/> instance originally passed in.
+        /// </returns>
+        public static AkkaConfigurationBuilder WithAzureDiscovery(
```
Aaronontheweb, line 98: So one downside of having ALL of the `.Hosting` plugins in the same NuGet package is that it's going to pull in all of their dependencies too - the K8s client, the AWS client, the Azure clients.... Do you think we'd be better off taking a dependency on Akka.Hosting directly inside the Akka.Discovery.X package so it's more modular?

Aaronontheweb (commented):
Have not reviewed API, but have some high level package management questions

SeanFarrow:
@Aaronontheweb
If we can take a dependency on Akka.Host from the Akka.Discovery.X packages, then it makes things easier to maintain.

src/hosting/Akka.Management.Hosting/AzureDiscoveryExtensions.cs
wesselkranenborg: I would vote for that. As a user I would not be happy with getting kubernetes and aws dependencies in my codebase when I'm deploying to azure web apps or container apps. It would not be a blocker to use though.

Aaronontheweb:
> If we can take a dependency on Akka.Host from the Akka.Discovery.X packages, then it makes things easier to maintain.

I agree

Arkatufus:
Ok, done, I've moved the extension methods to their packages instead

Aaronontheweb (approved):
LGTM

Aaronontheweb:
@Arkatufus We probably need to add some test coverage to these, but I'd recommend doing it in a separate PR.

PR #779, opened 2022-08-23
Make sure that exception is included in the log

Arkatufus:
Exception isn't being recorded in Akka.Management startup failure warning log


Aaronontheweb (approved):
LGTM

PR #780, opened 2022-08-23
Cleaner Akka.Discovery.Azure service shutdown

Arkatufus:
Make sure that when stop request is received that
- All children are stopped
- Further Lookup requests are ignored
- Multiple stop requests are ignored


Aaronontheweb (approved):
LGTM

PR #781, opened 2022-08-24
[K8S LEASE] Fix settings and settings spec, add Setup support

Arkatufus:
- Fix KubernetesSettings, `lease-operation-timeout` was not being used in code and the spec was not testing it correctly.
- Add Setup support for Akka.Hosting integration


Aaronontheweb (approved):
LGTM

PR #789, opened 2022-08-29
Harden Management and ClusterBootstrap interaction

Arkatufus:
## Changes
* AkkaManagement.Start() should be idempotent, it should start only once and returns the same Task on multiple invocation
* AkkaManagement should start automatically when added in `akka.extensions`, this behaviour should be consistent with other modules in Akka ecosystem
* AkkaManagement and ClusterBootstrap should co-exist when both are auto-started from Akka.Hosting in any combination

## Still a problem
* Without Akka.Hosting, ClusterBootstrap HTTP routes would not be registered if AkkaManagement starts *before* ClusterBootstrap starts, because ClusterBootstrap injects its routes manually when it starts. User would need to remember to add `.WithFallback(ClusterBootstrap.DefaultConfiguration())` to their HOCON configuration for this to work.

### DIFF
--- a/src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/AkkaHostingExtensions.cs
+++ b/src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/AkkaHostingExtensions.cs
@@ -5,9 +5,7 @@
 // -----------------------------------------------------------------------
 
 using System;
-using System.Linq;
 using Akka.Actor;
-using Akka.Configuration;
 using Akka.Hosting;
 
 namespace Akka.Management.Cluster.Bootstrap
@@ -54,6 +52,7 @@ namespace Akka.Management.Cluster.Bootstrap
             string serviceName = null,
             string serviceNamespace = null,
             string portName = null,
+            int? requiredContactPoints = null,
             bool? newClusterEnabled = null,
             bool autoStart = true)
             => builder.WithClusterBootstrap(new ClusterBootstrapSetup
@@ -63,7 +62,8 @@ namespace Akka.Management.Cluster.Bootstrap
                 {
                     ServiceName = serviceName,
                     ServiceNamespace = serviceNamespace,
-                    PortName = portName
+                    PortName = portName,
+                    RequiredContactPointsNr = requiredContactPoints
                 }
             }, autoStart);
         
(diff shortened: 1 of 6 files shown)


Aaronontheweb:
> ## Changes
> * AkkaManagement.Start() should be idempotent, it should start only once and returns the same Task on multiple invocation
> * AkkaManagement should start automatically when added in `akka.extensions`, this behaviour should be consistent with other modules in Akka ecosystem
> * AkkaManagement and ClusterBootstrap should co-exist when both are auto-started from Akka.Hosting in any combination
> 

100%

src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/AkkaHostingExtensions.cs
```diff
@@ -123,29 +123,14 @@ public static class AkkaHostingExtensions
...
-                        builder.AddHocon(config, HoconAddMode.Prepend);
-                    }
-                }
-            }
-            else
-            {
-                builder.AddHocon(
-                    "akka.management.http.routes.cluster-bootstrap = \"Akka.Management.Cluster.Bootstrap.ClusterBootstrapProvider, Akka.Management.Cluster.Bootstrap\"", 
-                    HoconAddMode.Prepend);
+                builder.WithExtensions(typeof(ClusterBootstrapProvider));
```
Aaronontheweb, line 126: nice - do we need to release a new version of Akka.Hosting so you can take advantage of the strongly typed version of this or does that not matter?

src/management/Akka.Management/Dsl/AkkaManagement.cs
```diff
@@ -53,6 +53,27 @@ public AkkaManagement(ExtendedActorSystem system)
...
                 });
+            
+            var autoStart = system.Settings.Config.GetStringList("akka.extensions")
+                .Any(s => s.Contains(typeof(AkkaManagementProvider).Name));
+            if (autoStart)
+            {
+                _log.Info("Akka.Management loaded through 'akka.extensions' auto starting bootstrap.");
+                // Akka Management hosts the HTTP routes used by bootstrap
+                // we can't let it block extension init, so run it in a different thread and let constructor complete
+                Task.Run(async () =>
```
Aaronontheweb, line 64: LGTM

src/management/Akka.Management/Dsl/AkkaManagement.cs
```diff
@@ -74,13 +95,22 @@ public AkkaManagement(ExtendedActorSystem system)
         /// <summary>
         /// Start an Akka HTTP server to serve the HTTP management endpoint.
         /// </summary>
-        public Task Start() => Start(Identity);
+        public Task<Uri> Start() => Start(Identity);
```
Aaronontheweb, line 98: LGTM

Aaronontheweb (approved):
LGTM

PR #791, opened 2022-08-30
Fix racy timeout tests

Arkatufus:
## Changes
- bumped the timeout value for ClusterBootstrapAutostartIntegrationSpec, timed out while it is still forming the cluster
- use `AwaitAssertAsync` around HTTP probing to make sure that the problem was not caused by Ceen slow socket setup


Aaronontheweb (approved):
LGTM

PR #802, opened 2022-09-01
[AWS.EC2] Add Settings and Setup class

Arkatufus:
## Breaking Change
Changed the `akka.discovery.aws-api-ec2-tag-based.credentials-provider` HOCON setting from a HOCON path to FQCN instead


Arkatufus:
Marked as draft because of the breaking change, wanted to make sure that this is OK

Aaronontheweb (approved):

Aaronontheweb (changes requested):
We should detect after we parse the change-breaking setting if the user is still using an old value - if so we should throw an exception and force them to update.

Arkatufus:
Done, added a verbose exception message and documented it in the code.

Aaronontheweb (approved):

Aaronontheweb (approved):
LGTM

PR #817, opened 2022-09-08
[Discovery.KubernetesApi] Add settings and setup classes

Arkatufus:


Arkatufus:
Used the wrong base branch when I created this. Recreating it.

PR #822, opened 2022-09-09
[Discovery.KubernetesApi] Add Akka.Hosting extension methods

Arkatufus:


src/discovery/kubernetes/Akka.Discovery.KubernetesApi/Akka.Discovery.KubernetesApi.csproj
```diff
@@ -10,6 +10,7 @@
 
     <ItemGroup>
         <PackageReference Include="Akka.Discovery" Version="$(AkkaVersion)" />
+        <PackageReference Include="Akka.Hosting" Version="0.4.2" />
```
Aaronontheweb, line 13: We should probably reference this from `Directory.Build.props` so all packages get the same version.

Aaronontheweb (changes requested):
One minor nitpick, otherwise LGTM

src/discovery/kubernetes/Akka.Discovery.KubernetesApi/Akka.Discovery.KubernetesApi.csproj
Arkatufus: Done

Aaronontheweb (approved):

PR #865, opened 2022-09-22
Add Akka.Coordination.Azure module

Arkatufus:


Aaronontheweb (changes requested):
@Arkatufus explained me during our review call why we went with writing lease data directly into a blob object rather than using Azure Blob Storage's native leasing capabilities - I found his explanation compelling but it needs to be written on this issue so we can refer back to it in the future in case this design is ever called into question by future Akka.NET users and contributors.

src/coordination/azure/Akka.Coordination.Azure.Tests/AzureApiSpec.cs
```diff
@@ -0,0 +1,106 @@
...
+using Xunit.Abstractions;
+
+namespace Akka.Coordination.Azure.Tests
+{
+    public class AzureApiSpec : TestKit.Xunit2.TestKit, IAsyncLifetime
+    {
+        private readonly AzureLeaseSettings _settings;
+        private readonly AzureApiImpl _underTest;
+        private const string LeaseName = "lease-1";
+        private const string ApiPath = "/apis/akka.io/v1/namespaces/lease/leases";
```
Aaronontheweb, line 27: Need to remove unused `const`s

src/coordination/azure/Akka.Coordination.Azure/IAzureApi.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// <summary>
+        /// Reads a Lease from the API server. If it doesn't exist it tries to create it.
+        /// The creation can fail due to another instance creating at the same time, in this case
+        /// the read is retried.
+        /// </summary>
+        /// <param name="name"></param>
+        /// <returns></returns>
+        Task<LeaseResource> ReadOrCreateLeaseResource(string name);
+
+        Task<Either<LeaseResource, LeaseResource>> UpdateLeaseResource(
```
Aaronontheweb, line 25: Need to add XML-DOC comment here

src/coordination/azure/Akka.Coordination.Azure/Internal/AzureApiImpl.cs
```diff
@@ -0,0 +1,338 @@
...
+            {
+                _log.Debug("Updating {0}", leaseName);
+                
+                var blobClient = _containerClient.Value.GetBlobClient(leaseName);
+                var leaseBody = new LeaseBody(ownerName, time);
+                var options = new BlobUploadOptions
+                {
+                    Conditions = new BlobRequestConditions
+                    {
+                        IfMatch = version
```
Aaronontheweb, line 98: So what's the importance of this? Make sure we can't override someone else's lease with our own?

Aaronontheweb (changes requested):
Design overall looks good - need a stress test and some documentation added.

src/coordination/azure/Akka.Coordination.Azure.Tests/AzureApiSpec.cs
Arkatufus: Done

src/coordination/azure/Akka.Coordination.Azure/IAzureApi.cs
Arkatufus: Done

src/coordination/azure/Akka.Coordination.Azure/Internal/AzureApiImpl.cs
Arkatufus: This is the concurrency lock, we can only update the lease if our ETag matches the server ETag, else the lease was taken by someone else

PR #909, opened 2022-10-05
Fix Azure storage, store only information provided

Arkatufus:
Store host name or address as null instead of trying to infer their value, because trying to infer this value can cause a lot of problem in certain systems (ie. systems that does not have DNS resolvers)


src/discovery/azure/Akka.Discovery.Azure/Actors/AzureDiscoveryGuardian.cs
```diff
@@ -136,25 +134,17 @@ private bool Initializing(object message)
...
 
-                            _host = Dns.GetHostName();
-                        }
-                        else
-                        {
-                            _host = _settings.HostName;
-                            var addresses = Dns.GetHostAddresses(_host);
-                            _address = addresses.First(i => i.AddressFamily == AddressFamily.InterNetwork && !Equals(i, IPAddress.Any));
-                        }
+                        _host = null;
```
Arkatufus, line 142: Just store null if host information was not given

src/discovery/azure/Akka.Discovery.Azure/Actors/AzureDiscoveryGuardian.cs
```diff
@@ -136,25 +134,17 @@ private bool Initializing(object message)
...
                     }
-                    catch (Exception ex)
+                    else
                     {
-                        throw new InvalidOperationException($"Failed to invoke Dns.GetHostEntry() for host [{_host}]", ex);
+                        _host = _settings.HostName;
+                        _address = null;
```
Arkatufus, line 147: Just store null if IP address was not given

Aaronontheweb:
@Arkatufus what motivated this PR? Did you run into this issue somewhere in the wild?

Arkatufus:
Yes, in an environment that doesn't support DNS lookup, trying to infer host name or trying to retrieve IP addresses of a host using DNS will fail

Arkatufus:
So it is a lot safer to just not do it, we don't need to have both IPAddress and host name to be populated for discovery.

src/discovery/azure/Akka.Discovery.Azure/Actors/AzureDiscoveryGuardian.cs
Aaronontheweb: LGTM

src/discovery/azure/Akka.Discovery.Azure/Actors/AzureDiscoveryGuardian.cs
Aaronontheweb: lgtm

Aaronontheweb (approved):
LGTM

PR #910, opened 2022-10-05
Fix IPv6 Uri string interpolation

Arkatufus:
IPv6 `IPAddress.ToString()` returns "2001:db8:85a3:8d3:1319:8a2e:370:7348", it needs to be bracketed ("[2001:db8:85a3:8d3:1319:8a2e:370:7348]") for use inside a URI string.


src/management/Akka.Management/Dsl/AkkaManagement.cs
```diff
@@ -179,10 +179,14 @@ private ManagementRouteProviderSettings ProviderSettings()
             const string protocol = "http"; // changed to "https" if ManagementRouteProviderSettings.withHttpsConnectionContext is use
 
             var basePath = !string.IsNullOrWhiteSpace(Settings.Http.BasePath) ? Settings.Http.BasePath + "/" : string.Empty;
-            var selfBaseUri = new Uri($"{protocol}://{Settings.Http.Hostname}:{Settings.Http.Port}{basePath}");
+            var hostName = IsIPv6(Settings.Http.Hostname) ? $"[{Settings.Http.Hostname}]" : Settings.Http.Hostname;
```
Arkatufus, line 182: The host name need to be "[2001:db8:85a3:8d3:1319:8a2e:370:7348]" or else `Uri` will throw an exception

Aaronontheweb (approved):
LGTM

PR #925, opened 2022-10-10
Fix ClusterBootstrap using wrong fallback port value

### LINKED ISSUE #924: Forming of an Akka.Cluster is not working anymore after upgrading to 0.3.0-beta1 of Akka.Discovery.Azure
**Version Information**
Akka.Discovery.Azure 0.3.0-beta1
Akka.Cluster 1.4.43

**Describe the bug**
The Akka.Cluster is not being formed.

**To Reproduce**
Have a working version of the code running with 0.2.5-beta4, upgrade to 0.3.0-beta1 without code changes and it will not form a cluster again (have tested it in Azure Container Apps)

**Expected behavior**
To form a cluster.

**Actual behavior**
In the logging I get the following logrules:

[22/10/10-03:21:28.8920][akka.tcp://service@10.250.1.19:12552/system/bootstrapCoordinator][akka://service/system/bootstrapCoordinator][0042]: Looking up [Lookup(service, , tcp)]
[22/10/10-03:21:28.8992][akka.tcp://service@10.250.1.19:12552/system/bootstrapCoordinator][akka://service/system/bootstrapCoordinator][0006]: Located service members based on: [Lookup(service, , tcp)]: [ResolvedTarget(, 18558, 10.250.0.200), ResolvedTarget(, 18558, 10.250.1.19), ResolvedTarget(, 18558, 10.250.1.38)], filtered to []
[22/10/10-03:21:28.9326][SelfAwareJoinDecider (akka://service)][akka://service/system/bootstrapCoordinator][0035]: Discovered [0] contact points, confirmed [0], which is less than the required [1], retrying
[22/10/10-03:21:29.0573][SelfAwareJoinDecider (akka://service)][akka://service/system/bootstrapCoordinator][0030]: Discovered [0] contact points, confirmed [0], which is less than the required [1], retrying
[22/10/10-03:21:29.3729][SelfAwareJoinDecider (akka://service)][akka://service/system/bootstrapCoordinator][0038]: Discovered [0] contact points, confirmed [0], which is less than the required [1], retrying

**Additional context**
After downgrading from 0.3.0-beta1 to 0.2.5-beta4 it started to work again.

Arkatufus:
Fixes #924

The fix isn't obvious at first.
The failure happened at this filtering method, it is filtering out all contact points from the result:
https://github.com/akkadotnet/Akka.Management/blob/28e6d297c532a58d1ac25797295f01114d2786e2/src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/Internal/BootstrapCoordinator.cs#L117-L136

There is nothing wrong with this method by itself, the problem is the fallback port passed into it. 

If `akka.management.cluster.bootstrap.contact-point.fallback-port` was not overriden, the default value is null or undefined. By default, the cluster bootstrap settings will try to fill this information by reading the `akka.management.http.port` HOCON value.
This by itself is not a problem, if we're using pure HOCON configuration. The `AkkaManagementSetup` class though, is a different story.

`AkkaManagementSetup` is being applied directly to the `AkkaManagementSettings` class, changing the settings class without touching the HOCON configuration. By this time, there are two management HTTP port value, the correct one is held inside `AkkaManagement.Settings` property, the other is the stale value inside `System.Settings.Config`. 

The bug happened because cluster bootstrap is using the stale value inside the HOCON config, and not the correct one in `AkkaManagement.Settings`.

### DIFF
--- a/src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/ClusterBootstrapSettings.cs
+++ b/src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/ClusterBootstrapSettings.cs
@@ -151,8 +151,7 @@ namespace Akka.Management.Cluster.Bootstrap
                 var contactPointConfig = config.GetConfig("akka.management.cluster.bootstrap.contact-point");
                 var fallback = contactPointConfig.GetString("fallback-port");
                 var fallbackPort = string.IsNullOrWhiteSpace(fallback) || fallback == "<fallback-port>"
-                    ? config.GetInt("akka.management.http.port")
-                    : int.Parse(fallback);
+                    ? (int?) null : int.Parse(fallback);
                 
                 return new ContactPointSettings(
                     fallbackPort: fallbackPort,
@@ -163,7 +162,7 @@ namespace Akka.Management.Cluster.Bootstrap
             }
             
             private ContactPointSettings(
-                int fallbackPort,
+                int? fallbackPort,
                 bool filterOnFallbackPort,
                 TimeSpan probingFailureTimeout,
                 TimeSpan probeInterval,
@@ -176,7 +175,7 @@ namespace Akka.Management.Cluster.Bootstrap
                 ProbeIntervalJitter = probeIntervalJitter;
             }
             
-            public int FallbackPort { get; }
+            public int? FallbackPort { get; }
             public bool FilterOnFallbackPort { get; }
             public TimeSpan ProbingFailureTimeout { get; }
             public TimeSpan ProbeInterval { get; }
--- a/src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/Internal/BootstrapCoordinator.cs
+++ b/src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/Internal/BootstrapCoordinator.cs
@@ -145,6 +145,7 @@ namespace Akka.Management.Cluster.Bootstrap.Internal
         private const string DecideTimerKey = "decide-key";
 
         private readonly Lookup _lookup;
+        private readonly int _fallbackPort;
 
         private ServiceContactsObservation _lastContactObservation;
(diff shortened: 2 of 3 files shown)


src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/ClusterBootstrapSettings.cs
```diff
@@ -151,8 +151,7 @@ public static ContactPointSettings Create(Config config)
                 var contactPointConfig = config.GetConfig("akka.management.cluster.bootstrap.contact-point");
                 var fallback = contactPointConfig.GetString("fallback-port");
                 var fallbackPort = string.IsNullOrWhiteSpace(fallback) || fallback == "<fallback-port>"
-                    ? config.GetInt("akka.management.http.port")
```
Arkatufus, line 154: Do not depend on HOCON config values of another plugin. The HOCON config is not the single source of truth for configuration anymore, since the final settings value can be influenced by a `Setup` class.

src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/Internal/BootstrapCoordinator.cs
```diff
@@ -164,6 +165,8 @@ public BootstrapCoordinator(ServiceDiscovery discovery, IJoinDecider joinDecider
                 _settings.ContactPointDiscovery.EffectiveName(Context.System),
                 _settings.ContactPointDiscovery.PortName,
                 _settings.ContactPointDiscovery.Protocol);
+
+            _fallbackPort = _settings.ContactPoint.FallbackPort ?? AkkaManagement.Get(Context.System).Settings.Http.Port;
```
Arkatufus, line 169: We read the actual settings value directly from the plugin, not the plugin HOCON config

Aaronontheweb (approved):
LGTM

PR #1152, opened 2022-12-28
Drop the janky scala routing and use Ceen built in routing instead

Arkatufus:


src/cluster.bootstrap/Akka.Management.Cluster.Bootstrap/ContactPoint/HttpBootstrapJsonProtocol.cs
```diff
@@ -41,14 +41,14 @@ public ClusterMember(Address node, long nodeUid, MemberStatus status, ImmutableH
         
         public sealed class SeedNodes
         {
-            public SeedNodes(Address selfNode, ImmutableHashSet<ClusterMember> nodes)
+            public SeedNodes(Address selfNode, ImmutableList<ClusterMember> nodes)
```
Aaronontheweb, line 44: Is it possible that this could allow duplicates?

Aaronontheweb: Nope, it can't - comes from the `Cluster` object itself which already de-dupes them.

src/management/Akka.Management/AkkaHostingExtensions.cs
```diff
@@ -48,21 +54,31 @@ public static class AkkaHostingExtensions
         /// </returns>
         public static AkkaConfigurationBuilder WithAkkaManagement(
             this AkkaConfigurationBuilder builder,
+            Dictionary<string, IManagementRouteProvider>? providers = null,
```
Aaronontheweb, line 57: Move this to the back of the parameter list, since it's an advanced feature

Aaronontheweb: I'd make Akka.Cluster.Bootstrap the default here

Aaronontheweb (approved):
LGTM

PR #1153, opened 2022-12-29
Merge Cluster.Bootstrap package into Akka.Management

Arkatufus:
## Changes

- Merge Akka.Management.Cluster.Bootstrap into Akka.Management package, replacing the default health check endpoint.
- Cluster.Bootstrap "/bootstrap/seed-nodes" endpoint will return `ServiceUnavailable` HTTP code if akka.cluster is not available.
- Cluster.Bootstrap will exit immediately if `Start()`ed when akka.cluster is not available.

## Breaking Change
Since the assembly name is changed, all FQCN string reference need to be changed from "...ClusterBootstrapProvider, Akka.Management.Cluster.Bootstrap" to "...ClusterBootstrapProvider, Akka.Management". This is an unavoidable change.


Aaronontheweb:
So do we need to deprecate the separate Akka.Cluster.Boostrap package? And what will we need to do with the documentation going forward?

Arkatufus:
The old documentations are merged with the main Akka.Management documentation

Arkatufus:
We'd have to deprecate the old package wether we like it or not, i think...

PR #1181, opened 2023-01-05
Change Azure lease hosting extensions to use options class instead of setup

Arkatufus:
## Changes

Change `WithAzureLease()` to accept `AzureLeaseOption` instead of `AzureLeaseSetup` to make it consistent with other `Akka.Hosting` extension methods in the ecosystem


Aaronontheweb (approved):
LGTM

PR #1200, opened 2023-01-10
Change Discovery.Kubernetes extension method to take options instead of setup

Arkatufus:
Change extension method argument to take options instead of setup class


Aaronontheweb (approved):
LGTM

Issue #1251, opened 2023-01-26
[Lease.Azure] Unexpected cluster instability


Arkatufus:
**Version Information**
Version of Akka.NET? 1.4.48
Which Akka.NET Modules? Akka.Coordination.Azure 1.0.0-beta2

**Describe the bug**

After upgrading to these module versions:
<img width="1300" alt="Versions" src="https://user-images.githubusercontent.com/8293237/214854224-1a286a0f-46c2-4e39-b1b6-5fa6bba7439b.png">

Sharded cluster were unstable because the lease state were corrupted:
<img width="927" alt="Exceptions" src="https://user-images.githubusercontent.com/8293237/214854645-0ac5e439-ba0d-4b0e-a6f0-a0c97e60a97b.png">

**Expected behavior**
Sharded cluster should be stable, lease should work as intended.

**Actual behavior**
Sharded cluster were unstable

wesselkranenborg:
Can the issue be that I'm still using `<PackageReference Include="Akka.Management.Cluster.Bootstrap" Version="0.3.0-beta4" />`? See that there is no 1.0.0 version of that is but that there is now an `Akka.Management` 1.0.0.

Arkatufus:
That could be the problem, yes. Please let us know if that fixes the problem.

wesselkranenborg:
First impression is that the cluster is stable for now (but the issues back then only occurred after x hours) but there are still errors in the logs. It are errors with this blob file:
> https://<storageaccountname>.blob.core.windows.net/akka-coordination-lease/system-state-service-singleton-akkasystem-state-servicesystemsh

<img width="782" alt="image" src="https://user-images.githubusercontent.com/4964604/215411259-f2f7e93a-cbf0-49a6-adbb-7b3b1c8abab8.png">

The errors are also pretty consistent, these are the failed dependency calls from the last 12 hrs:
<img width="780" alt="image" src="https://user-images.githubusercontent.com/4964604/215411633-db8ecfb3-572d-4ba8-90af-6de3d8eb9326.png">


These errors started to happen after the upgrade. In Akka.Coordination.Azure `0.3.0-beta4` (we still use that in production) these errors are not shown in the logs.

Arkatufus:
We'll look into this, thanks!

wesselkranenborg:
Good find! So basically it was only noise in the system which is gone in version 1.0.1?

Arkatufus:
[...]

PR #1289, opened 2023-02-06
Fix Lease release/acquire operation logic

Arkatufus:
Posible Fix for #1251

## Changes
* Lease release op when lease not acquired is a no-op
* Lease acquire when lease was being acquired returns idempotent operation task instead of timing out

### DIFF
--- a/src/coordination/azure/Akka.Coordination.Azure/AzureLease.cs
+++ b/src/coordination/azure/Akka.Coordination.Azure/AzureLease.cs
@@ -102,8 +102,8 @@ namespace Akka.Coordination.Azure
                         case LeaseActor.LeaseReleased:
                             return true;
                         case LeaseActor.InvalidReleaseRequest:
-                            _log.Debug("Tried to release a lease that is not acquired");
-                            return false;
+                            _log.Info("Tried to release a lease that is not acquired");
+                            return true;
                         case Status.Failure f:
                             throw new LeaseException($"Failure while releasing lease: {f.Cause.Message}", f.Cause);
                         default:
--- a/src/coordination/azure/Akka.Coordination.Azure/AzureLease.cs
+++ b/src/coordination/azure/Akka.Coordination.Azure/AzureLease.cs
@@ -5,8 +5,11 @@
 // -----------------------------------------------------------------------
 
 using System;
+using System.Linq;
+using System.Runtime.ExceptionServices;
 using System.Text;
 using System.Text.RegularExpressions;
+using System.Threading;
 using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Configuration;
@@ -45,6 +48,8 @@ namespace Akka.Coordination.Azure
         private readonly TimeSpan _timeout;
         private readonly string _leaseName;
         private readonly IActorRef _leaseActor;
+        private readonly object _acquireLock = new ();
+        private Task<bool>? _acquireTask;
 
         public AzureLease(LeaseSettings settings, ExtendedActorSystem system) :
             this(system, new AtomicBoolean(), settings)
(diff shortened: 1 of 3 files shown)


src/coordination/azure/Akka.Coordination.Azure/AzureLease.cs
```diff
@@ -109,25 +120,59 @@ public override async Task<bool> Release()
...
+        public override Task<bool> Acquire(Action<Exception?>? leaseLostCallback)
         {
-            try
+            lock (_acquireLock)
             {
+                if (_acquireTask is not null)
+                {
+                    if(_log.IsDebugEnabled)
+                        _log.Debug("Lease is already being acquired");
+                    return _acquireTask;
```
Aaronontheweb, line 131: LGTM

src/coordination/azure/Akka.Coordination.Azure/AzureLease.cs
```diff
@@ -92,12 +97,18 @@ public override async Task<bool> Release()
...
                 {
-                    LeaseActor.LeaseReleased _ => true,
-                    LeaseActor.InvalidRequest req => throw new LeaseException(req.Reason),
-                    _ => throw new LeaseException($"Unexpected response type: {result.GetType()}")
-                };
+                        case LeaseActor.LeaseReleased:
+                            return true;
+                        case LeaseActor.InvalidReleaseRequest:
+                            _log.Info("Tried to release a lease that is not acquired");
+                            return true;
```
Arkatufus, line 106: Logically, since we didn't have the lease, we should return true.
This will cause the plugins to report that the lease have been successfully released, which can be confusing.
This is why the log should be emitted in info level so that we can track that the lease was "successfully released" because we didn't have it in the first place.

src/coordination/azure/Akka.Coordination.Azure/AzureLease.cs
```diff
@@ -92,12 +97,18 @@ public override async Task<bool> Release()
...
+                switch (result)
                 {
-                    LeaseActor.LeaseReleased _ => true,
-                    LeaseActor.InvalidRequest req => throw new LeaseException(req.Reason),
-                    _ => throw new LeaseException($"Unexpected response type: {result.GetType()}")
-                };
+                        case LeaseActor.LeaseReleased:
+                            return true;
+                        case LeaseActor.InvalidReleaseRequest:
+                            _log.Info("Tried to release a lease that is not acquired");
```
Aaronontheweb, line 105: I'd mark this as `DEBUG`

src/coordination/azure/Akka.Coordination.Azure/AzureLease.cs
```diff
@@ -109,25 +120,59 @@ public override async Task<bool> Release()
         public override Task<bool> Acquire()
             => Acquire(null);
 
-        public override async Task<bool> Acquire(Action<Exception?>? leaseLostCallback)
+        public override Task<bool> Acquire(Action<Exception?>? leaseLostCallback)
```
Arkatufus, line 123: To make this idempotent, we could not use async...await pattern

src/coordination/azure/Akka.Coordination.Azure/AzureLease.cs
```diff
@@ -109,25 +120,59 @@ public override async Task<bool> Release()
...
-                    LeaseActor.InvalidRequest req => throw new LeaseException(req.Reason),
-                    _ => throw new LeaseException($"Unexpected response type: {result.GetType()}")
-                };
-            }
-            catch (AskTimeoutException)
-            {
-                throw new LeaseTimeoutException(
-                    $"Timed out trying to acquire lease [{_leaseName}, {_settings.OwnerName}]. It may still be taken.");
+                _acquireTask = _leaseActor.Ask(new LeaseActor.Acquire(leaseLostCallback), _timeout)
+                    .ContinueWith(t =>
```
Aaronontheweb, line 137: LGTM

Aaronontheweb (approved):
LGTM

src/coordination/azure/Akka.Coordination.Azure/AzureLease.cs
Arkatufus: it needs to be `INFO` because if its not, we could not debug if a lease was "released" because it was actually released or because we didn't have it in the first place

PR #1310, opened 2023-02-13
[Coordination.Azure] Fix lease update and prune bug

### LINKED ISSUE #1251: [Lease.Azure] Unexpected cluster instability

Arkatufus:
Fixes #1251

## Changes

* Force "void" `PipeTo` to return `Status.Success` on successful invocation.

### DIFF
--- a/src/discovery/azure/Akka.Discovery.Azure/Actors/HeartbeatActor.cs
+++ b/src/discovery/azure/Akka.Discovery.Azure/Actors/HeartbeatActor.cs
@@ -74,7 +74,7 @@ namespace Akka.Discovery.Azure.Actors
                     if(_log.IsDebugEnabled)
                         _log.Debug("Updating cluster member entry TTL");
 
-                    ExecuteUpdateOpWithRetry().PipeTo(Self);
+                    ExecuteUpdateOpWithRetry().PipeTo(Self, success: () => Status.Success.Instance);
                     break;
                 
                 case Status.Success _:
@@ -89,7 +89,7 @@ namespace Akka.Discovery.Azure.Actors
                     }
                     
                     _log.Warning(f.Cause, "Failed to update TTL heartbeat, retrying");
-                    ExecuteUpdateOpWithRetry().PipeTo(Self);
+                    ExecuteUpdateOpWithRetry().PipeTo(Self, success: () => Status.Success.Instance);
                     break;
                 
                 default:
--- a/src/discovery/azure/Akka.Discovery.Azure/Actors/HeartbeatActor.cs
+++ b/src/discovery/azure/Akka.Discovery.Azure/Actors/HeartbeatActor.cs
@@ -74,7 +74,7 @@ namespace Akka.Discovery.Azure.Actors
                     if(_log.IsDebugEnabled)
                         _log.Debug("Updating cluster member entry TTL");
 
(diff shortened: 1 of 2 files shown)


src/discovery/azure/Akka.Discovery.Azure/Actors/HeartbeatActor.cs
```diff
@@ -74,7 +74,7 @@ protected override void OnReceive(object message)
                     if(_log.IsDebugEnabled)
                         _log.Debug("Updating cluster member entry TTL");
 
-                    ExecuteUpdateOpWithRetry().PipeTo(Self);
+                    ExecuteUpdateOpWithRetry().PipeTo(Self, success: () => Status.Success.Instance);
```
Aaronontheweb: What happens on failure?

Aaronontheweb (commented):
Tests are failing, but also left a question

src/discovery/azure/Akka.Discovery.Azure/Actors/HeartbeatActor.cs
Arkatufus: `PipeTo` always emits a `Status.Failure` on exceptions, unless you override it with a failure callback. The problem is with `PipeTo` that does not return anything (`async Task`, not `async Task<T>`), they're more akin to starting a detached `Task` that returns a `Status.Failure` when an exception was thrown

Aaronontheweb: That would do it. Got it, makes sense.

Aaronontheweb (approved):
LGTM

Aaronontheweb:
```
   at Akka.Coordination.Azure.Internal.AzureApiImpl.LeaseResourceExists(String leaseName) in /home/vsts/work/1/s/src/coordination/azure/Akka.Coordination.Azure/Internal/AzureApiImpl.cs:line 227
   at Akka.Coordination.Azure.Internal.AzureApiImpl.ReadOrCreateLeaseResource(String name) in /home/vsts/work/1/s/src/coordination/azure/Akka.Coordination.Azure/Internal/AzureApiImpl.cs:line 59
   at Akka.Coordination.Azure.Tests.AzureApiSpec.ShouldUpdateLeaseConflict() in /home/vsts/work/1/s/src/coordination/azure/Akka.Coordination.Azure.Tests/AzureApiSpec.cs:line 80
--- End of stack trace from previous location where exception was thrown ---
----- Inner Stack Trace -----
```

Anything of note going on with this test failure @Arkatufus ?

Arkatufus:
I think its the linux CI/CD image socket networking problem, the default connection timeout is 1 second, which I think a bit problematic on AzDo

Arkatufus:
@Aaronontheweb https://github.com/akkadotnet/Akka.Management/pull/1315 should fix this, it bumps the default timeout value

PR #2475, opened 2024-04-17
Fix Coordination.Kubernetes settings

Arkatufus:
Coordination settings is handled like persistence config where the actual config is composed by the plugin provider and it is a merged config of the actual config and a default config.

`KubernetesSettings` shouldn't base its config on `ActorSystem.Settings`, it should base its config on the `LeaseSettings.LeaseConfig` which is the actual composed config for the lease.


Aaronontheweb (approved):
LGTM

PR #2476, opened 2024-04-17
Harden Akka.Management IPV6 address detection

Arkatufus:
Akka.Management IPv6 detection is very simplistic, this hardens the detection.


Aaronontheweb (approved):
LGTM

PR #2557, opened 2024-06-11
Fix CVE-2024-21319 and CVE-2022-26907

Arkatufus:
CVE-2024-21319 and CVE-2022-26907 only applies to all `KubernetesClient` related projects that targets .NET Standard 2.0. It is very unlikely that there will be any update for this version.

## Changes

Reference clean version of `Microsoft.IdentityModel.JsonWebTokens` and `System.IdentityModel.Tokens.Jwt` directly, only if the project targets netstandard2.0


Aaronontheweb (approved):
LGTM

PR #2595, opened 2024-06-25
Update Discovery.Azure to support multi-config

Arkatufus:
## Changes

* Add ability to make the plugin read-only. A read-only plugin does not participate in updating the Azure table entries.
* Add Akka.Hosting support
* Add multi-config support
  * Make discovery plugin id configurable (was hardcoded)
  * Automatically generate default config fallback for each named discovery id
  * Only plugin with IsDefaultPlugin set to true will be used as `akka.discovery.method` value

### DIFF
--- a/src/discovery/azure/Akka.Discovery.Azure.Tests/AzureDiscoverySettingsSpecs.cs
+++ b/src/discovery/azure/Akka.Discovery.Azure.Tests/AzureDiscoverySettingsSpecs.cs
@@ -25,7 +25,8 @@ namespace Akka.Discovery.Azure.Tests
             var assemblyName = typeof(AzureServiceDiscovery).Assembly.FullName!.Split(',')[0].Trim();
             var config = AzureServiceDiscovery.DefaultConfig.GetConfig("akka.discovery.azure");
             config.GetString("class").Should().Be($"{typeof(AzureServiceDiscovery).Namespace}.{nameof(AzureServiceDiscovery)}, {assemblyName}");
-            
+
+            settings.ReadOnly.Should().BeFalse();
             settings.ServiceName.Should().Be("default");
             settings.HostName.Should().Be(Dns.GetHostName());
             settings.Port.Should().Be(8558);
@@ -46,6 +47,7 @@ namespace Akka.Discovery.Azure.Tests
             var settings = AzureDiscoverySettings.Create(AzureServiceDiscovery.DefaultConfig);
             var empty = AzureDiscoverySettings.Empty;
 
+            empty.ReadOnly.Should().Be(settings.ReadOnly);
             empty.ServiceName.Should().Be(settings.ServiceName);
             empty.HostName.Should().Be(settings.HostName);
             empty.Port.Should().Be(settings.Port);
@@ -66,6 +68,7 @@ namespace Akka.Discovery.Azure.Tests
             var uri = new Uri("https://whatever.com");
             var credential = new DefaultAzureCredential();
             var settings = AzureDiscoverySettings.Empty
+                .WithReadOnlyMode(true)
                 .WithServiceName("a")
                 .WithPublicHostName("host")
                 .WithPublicPort(1234)
@@ -76,7 +79,8 @@ namespace Akka.Discovery.Azure.Tests
                 .WithPruneInterval(3.Seconds())
                 .WithOperationTimeout(4.Seconds())
                 .WithAzureCredential(uri, credential);
-            
+
+            settings.ReadOnly.Should().BeTrue();
             settings.ServiceName.Should().Be("a");
             settings.HostName.Should().Be("host");
             settings.Port.Should().Be(1234);
@@ -97,6 +101,7 @@ namespace Akka.Discovery.Azure.Tests
             var uri = new Uri("https://whatever.com");
             var credential = new DefaultAzureCredential();
             var setup = new AzureDiscoverySetup()
+                .WithReadOnlyMode(true)
                 .WithServiceName("a")
                 .WithPublicHostName("host")
                 .WithPublicPort(1234)
@@ -109,7 +114,8 @@ namespace Akka.Discovery.Azure.Tests
                 .WithAzureCredential(uri, credential);
             
             var settings = setup.Apply(AzureDiscoverySettings.Empty);
-            
+
+            settings.ReadOnly.Should().BeTrue();
             settings.ServiceName.Should().Be("a");
             settings.HostName.Should().Be("host");
             settings.Port.Should().Be(1234);
--- a/src/discovery/azure/Akka.Discovery.Azure/Actors/AzureDiscoveryGuardian.cs
+++ b/src/discovery/azure/Akka.Discovery.Azure/Actors/AzureDiscoveryGuardian.cs
@@ -77,7 +77,8 @@ namespace Akka.Discovery.Azure.Actors
         private IPAddress? _address;
         private readonly int _port;
         private readonly CancellationTokenSource _shutdownCts;
-        
+
+        private readonly bool _readOnly;
         private readonly TimeSpan _backoff;
         private readonly TimeSpan _maxBackoff;
         private int _retryCount;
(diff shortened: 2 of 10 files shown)


src/discovery/azure/Akka.Discovery.Azure.Tests/AzureDiscoverySettingsSpecs.cs
```diff
@@ -66,6 +68,7 @@ public void SettingsWithOverrideTest()
             var uri = new Uri("https://whatever.com");
             var credential = new DefaultAzureCredential();
             var settings = AzureDiscoverySettings.Empty
+                .WithReadOnlyMode(true)
```
Aaronontheweb, line 71: This makes sense - there haven't been any use cases before now where we needed read-only.

src/discovery/azure/Akka.Discovery.Azure/Actors/AzureDiscoveryGuardian.cs
```diff
@@ -244,13 +250,21 @@ private bool Running(object message)
...
                         Context.Stop(child);
-                    
-                    var sender = Sender;
-                    Client.RemoveSelf(_shutdownCts.Token)
-                        .PipeTo(Self, 
-                            success: () => new DiscoveryStopped(sender), 
-                            failure: e => new DiscoveryStopFailed(sender, e));
-                    Become(Stopping);
+
+                    if (!_readOnly)
```
Aaronontheweb, line 254: LGTM

src/discovery/azure/Akka.Discovery.Azure/AkkaHostingExtensions.cs
```diff
@@ -123,11 +128,17 @@ public static class AkkaHostingExtensions
             TableClientOptions? tableClientOptions = null,
             string? serviceName = null,
             string? publicHostname = null,
-            int? publicPort = null)
+            int? publicPort = null,
+            string discoveryId = "azure",
```
Aaronontheweb: What does `discoveryId` mean?

src/discovery/azure/Akka.Discovery.Azure/Extensions.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Linq;
+using Akka.Configuration.Hocon;
+using Akka.Util.Internal;
+
+namespace Akka.Discovery.Azure;
+
+public static class Extensions
+{
+    public static Configuration.Config MoveTo(this Configuration.Config config, string path)
```
Aaronontheweb: Can this be made `internal`?

src/discovery/azure/Akka.Discovery.Azure/AkkaHostingExtensions.cs
```diff
@@ -123,11 +128,17 @@ public static class AkkaHostingExtensions
...
+            bool? readOnly = null,
+            bool isDefaultPlugin = true)
```
Aaronontheweb, line 154: An idea: what if we just used the `service-name` to alias the HOCON settings, so there's only one moving thing that needs to be tracked? This way, we don't even need to worry about default plugins et al.

Aaronontheweb (commented):
Left some thoughts

src/discovery/azure/Akka.Discovery.Azure/AkkaHostingExtensions.cs
Arkatufus: Discovery Id is the HOCON configuration key the final HOCON settings will live in. For example, "azure" the default Id will be placed in "akka.discovery.azure", just like how Akka.Persistence journal Id works. This will allow it to be addressed and loaded by the Akka.Discovery plugin later on.

src/discovery/azure/Akka.Discovery.Azure/Extensions.cs
Arkatufus: Will do

src/discovery/azure/Akka.Discovery.Azure/AkkaHostingExtensions.cs
Arkatufus: That works, but that will break previous behavior.

Aaronontheweb: That would probably be ok as long as we had some way of covering the default case from before

src/discovery/azure/Akka.Discovery.Azure/Extensions.cs
Arkatufus: done

src/discovery/azure/Akka.Discovery.Azure/AzureServiceDiscovery.cs
```diff
@@ -28,13 +28,19 @@ public class AzureServiceDiscovery : ServiceDiscovery
...
         {
             _system = system;
             _log = Logging.GetLogger(system, typeof(AzureServiceDiscovery));
             
-            _system.Settings.InjectTopLevelFallback(DefaultConfig);
-            _settings = AzureDiscoverySettings.Create(system);
+            var fullConfig = config.WithFallback(DefaultConfig.GetConfig("akka.discovery.azure"));
+            _settings = AzureDiscoverySettings.Create(system, fullConfig);
             
             var setup = _system.Settings.Setup.Get<AzureDiscoverySetup>();
```
Arkatufus, line 48: @Aaronontheweb This line is currently problematic, a single registered `AzureDiscoverySetup` instance can pollute every instance of this ServiceDiscovery instance. Would need to look at how Akka.Persistence handles this.

Arkatufus: Fixed this using a hack by embedding the discovery id inside the Akka.Hosting generated HOCON config. We can map/corellate which config goes to which setup class this way.

src/discovery/azure/Akka.Discovery.Azure/AkkaDiscoveryOptions.cs
```diff
@@ -39,9 +41,16 @@ public class AkkaDiscoveryOptions: IHoconOption
...
     {
         var sb = new StringBuilder();
-        sb.AppendLine($"{FullPath} {{");
+        sb.AppendLine($"{AzureServiceDiscovery.FullPath(ConfigPath)} {{");
         sb.AppendLine($"class = {Class.AssemblyQualifiedName!.ToHocon()}");
 
+        // We're going to cheat and embed the config path/discovery id here
+        // because we need to correlate the setup files with the config
+        // during run-time
+        sb.AppendLine($"discovery-id = {ConfigPath}");
```
Aaronontheweb, line 50: LGTM

src/discovery/azure/Akka.Discovery.Azure/AzureDiscoveryMultiSetup.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Collections.Immutable;
+using Akka.Actor.Setup;
+
+namespace Akka.Discovery.Azure;
+
+internal sealed class AzureDiscoveryMultiSetup : Setup
```
Aaronontheweb, line 6: I assume this is internal because it's only accessible via Akka.Hosting

src/discovery/azure/Akka.Discovery.Azure/reference.conf
```diff
@@ -8,6 +8,11 @@ akka.discovery {
   azure {
     class = "Akka.Discovery.Azure.AzureServiceDiscovery, Akka.Discovery.Azure"
 
+    # If set to true, the extension will not participate in updating the Azure table.
+    # Only need to be set to true if the extension is being used by a read only extension 
+    # such as ClusterClient contact discovery.
+    read-only = false
```
Aaronontheweb, line 14: LGTM

Aaronontheweb (approved):
For a real release we'll need to add documentation for these features, but since we're beta-testing them I'm fine shipping as is

PR #2604, opened 2024-06-28
Update Discovery.Config to support multi-config

Arkatufus:
## Changes

* Make Discovery.Config.Hosting compatible with multi-config


Aaronontheweb (commented):
LGTM

PR #2614, opened 2024-07-02
Add ClusterClientReceptionist actor path and Remote address endpoint

Arkatufus:
## Changes

* Add ClusterClientReceptionist ActorPath HTTP query endpoint at "/cluster-client/receptionist"
* Add Akka.Remote Address HTTP query endpoint at "/remote/address/"
* Both endpoint routes are loaded by `Akka.Management` by default.

### DIFF
--- a/src/management/Akka.Management.Tests/AkkaManagementSettingsSpec.cs
+++ b/src/management/Akka.Management.Tests/AkkaManagementSettingsSpec.cs
@@ -39,9 +39,17 @@ namespace Akka.Management.Tests
             http.EffectiveBindHostname.Should().Be(defaultHostname);
             http.EffectiveBindPort.Should().Be(8558);
             http.BasePath.Should().BeEmpty();
-            http.RouteProviders.Count.Should().Be(1);
+            http.RouteProviders.Count.Should().Be(3);
+            
             http.RouteProviders[0].Name.Should().Be("cluster-bootstrap");
             http.RouteProviders[0].FullyQualifiedClassName.Should().Be("Akka.Management.Cluster.Bootstrap.ClusterBootstrapProvider, Akka.Management");
+            
+            http.RouteProviders[1].Name.Should().Be("remote-address");
+            http.RouteProviders[1].FullyQualifiedClassName.Should().Be("Akka.Management.Routes.AddressRouteProvider, Akka.Management");
+            
+            http.RouteProviders[2].Name.Should().Be("cluster-client-receptionist");
+            http.RouteProviders[2].FullyQualifiedClassName.Should().Be("Akka.Management.Routes.ClusterClientReceptionistRouteProvider, Akka.Management");
+            
             http.RouteProvidersReadOnly.Should().BeTrue();
         }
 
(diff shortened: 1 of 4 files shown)


src/management/Akka.Management/Resources/reference.conf
```diff
@@ -59,6 +59,10 @@ akka.management {
     routes {
         # registers bootstrap routes to be included in akka-management's http endpoint
         cluster-bootstrap = "Akka.Management.Cluster.Bootstrap.ClusterBootstrapProvider, Akka.Management"
+        # register remoting address routes to be included in akka-management's http endpoint
+        remote-address = "Akka.Management.Routes.AddressRouteProvider, Akka.Management"
+        # register ClusterClientReceptionist actor path routes to be included in akka-management's http endpoint
+        cluster-client-receptionist = "Akka.Management.Routes.ClusterClientReceptionistRouteProvider, Akka.Management"
```
Arkatufus, line 65: Both new endpoints are loaded by default when `Akka.Management` is started

src/management/Akka.Management/Routes/ClusterClientReceptionistRouteProvider.cs
```diff
@@ -0,0 +1,91 @@
...
+            });
+            await context.HttpContext.Response.WriteAllJsonAsync(response);
+            return true;
+        }
+        
+        // Check that ClusterClientReceptionist is available
+        var config = context.ActorSystem.Settings.Config.GetConfig("akka.cluster.client.receptionist");
+        if(config is null)
+        {
+            context.HttpContext.Response.StatusCode = HttpStatusCode.ServiceUnavailable;
```
Arkatufus, line 70: Return a 503 - Service Unavailable if `akka.cluster.client.receptionist` HOCON config is not available. 
We're not referencing Akka.Cluster.Tools NuGet, no type checking is nescessary, it is the ClusterClient job to check that this actor path is actually alive.

Arkatufus (commented):
Self review

src/management/Akka.Management/Routes/ClusterClientReceptionistRouteProvider.cs
Aaronontheweb: If the default HOCON from the `ClusterClient` package gets loaded but there's no actual receptionist running, what happens then?

src/management/Akka.Management.Tests/AkkaManagementSettingsSpec.cs
Aaronontheweb, line 1: LGTM

src/management/Akka.Management/Routes/AddressRouteProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+{
+    public Route[] Routes => new Route[] { ("/remote/address", this) };
+    
+    public override async Task<bool> HandleAsync(IAkkaHttpContext context)
+    {
+        if (context.HttpContext.Request.Method.ToLowerInvariant() != "get")
+            return false;
+        
+        // Check that remoting is in effect
+        if (((ExtendedActorSystem)context.ActorSystem).Provider is not IRemoteActorRefProvider actorProvider)
```
Aaronontheweb, line 30: lol, I get why this is necessary but it would be pretty funny if someone managed to get Akka.Management up and running without clustering or remoting.

src/management/Akka.Management/Routes/ClusterClientReceptionistRouteProvider.cs
Aaronontheweb: Rephrasing my question: is loading the default HOCON from the ClusterClient module enough to flip this bit? Or does the user have to actively populate it? The correct answers must be 1. "no" 2. "yes"

Aaronontheweb (commented):
Left some questions

src/management/Akka.Management/Routes/ClusterClientReceptionistRouteProvider.cs
Arkatufus: This is strictly `Akka.Management`, using `ClusterClient` does not start `Akka.Management`.

For the discovery feature to work, the user has to start `Akka.Management` on the `ClusterClientReceptionist` side.

```csharp
builder
  .WithAkkaManagement() // <-- This MUST be used
  .WithClusterClientReceptionist(); // <-- This DOES NOT load Akka.Management
```

```csharp
builder
  .WithClusterClient<Key>(); // <-- This definitely DOES NOT load Akka.Management, its on the wrong side of the connection.
```

Aaronontheweb (approved):
LGTM

PR #2615, opened 2024-07-02
[Management] Harden ClusterClientReceptionist actor path endpoint

Arkatufus:
## Changes

Harden ClusterClientReceptionist ActorPath endpoint validation to also validate the receptionist name. Returns 500 - Internal Server Error if receptionist name is null, empty, whitespace, or is an invalid actor name


Aaronontheweb (approved):
LGTM

PR #2636, opened 2024-07-10
Fix seed-nodes endpoint used by ClusterBootstrap

Arkatufus:
## Changes

Make the "bootstrap/seed-nodes" endpoint return an empty seed node list if it is downed, exiting, leaving, or removed from the cluster.


Aaronontheweb (approved):
LGTM - this should help keep garbage out of the bootstrapping process from nodes that are on their way out the door.

Arkatufus:
Returning a non-200 HTTP status broke the spec, it needs to return an empty seed node list.

Aaronontheweb (approved):
LGTM

PR #2651, opened 2024-07-15
Add Discovery.AwsApi multi-config support

Arkatufus:
## Changes

Add multi-config support to Akka.Discovery.AwsApi


Aaronontheweb (approved):
LGTM

PR #2653, opened 2024-07-15
Fix missing Discovery.Azure IExtension

Arkatufus:
Discovery.Azure is missing `AzureDiscovery` class that extends IExtension

## Changes

Add `AzureDiscovery` class

### DIFF
--- a/src/discovery/azure/Akka.Discovery.Azure/AkkaHostingExtensions.cs
+++ b/src/discovery/azure/Akka.Discovery.Azure/AkkaHostingExtensions.cs
@@ -240,8 +240,12 @@ namespace Akka.Discovery.Azure
             AkkaDiscoveryOptions options)
         {
             options.Apply(builder);
-
-            builder.AddHocon(AzureServiceDiscovery.DefaultConfig, HoconAddMode.Append);
+            
+            // force start the module
+            builder.AddStartup((system, registry) =>
+            {
+                AzureDiscovery.Get(system);
+            });
             return builder;
         }
     }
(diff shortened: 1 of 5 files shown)


src/discovery/azure/Akka.Discovery.Azure/AkkaHostingExtensions.cs
```diff
@@ -240,8 +240,12 @@ public static class AkkaHostingExtensions
             AkkaDiscoveryOptions options)
         {
             options.Apply(builder);
-
-            builder.AddHocon(AzureServiceDiscovery.DefaultConfig, HoconAddMode.Append);
+            
+            // force start the module
+            builder.AddStartup((system, registry) =>
+            {
+                AzureDiscovery.Get(system);
```
Aaronontheweb, line 247: Why is this necessary now but not before?

Aaronontheweb (commented):
Just left a question

src/discovery/azure/Akka.Discovery.Azure/AkkaHostingExtensions.cs
Arkatufus: Before, the discovery module was automatically started by ClusterBootstrap during startup. Since ClusterBootstrap might not run in tandem with AzureDiscovery anymore, we need to force start it.

Aaronontheweb: Got it, makes sense

Aaronontheweb (approved):
LGTM

PR #2706, opened 2024-08-15
Fix AWS ECS Discovery

Arkatufus:
AWS ECS tag matching logic is wrong, it assumes that all of the tags have to match exactly

## Changes

* Change the tag matching logic to match what Scala Akka is doing
* Make sure that the `Diff()` logic follows Scala `list[T].diff()` behavior exactly

### DIFF
--- a/src/discovery/aws/Akka.Discovery.AwsApi/Ecs/EcsServiceDiscovery.cs
+++ b/src/discovery/aws/Akka.Discovery.AwsApi/Ecs/EcsServiceDiscovery.cs
@@ -24,6 +24,7 @@ namespace Akka.Discovery.AwsApi.Ecs
         public static readonly EcsTagComparer TagComparer = new ();
         
         private readonly EcsServiceDiscoverySettings _settings;
+        private readonly ImmutableList<AwsTag> _tags;
 
         private AmazonECSClient? _clientDoNotUseDirectly;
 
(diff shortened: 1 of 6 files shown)


src/discovery/aws/Akka.Discovery.AwsApi/Ecs/EcsServiceDiscovery.cs
```diff
@@ -82,13 +84,13 @@ private static async Task<List<EcsTask>> ResolveTasks(
...
             string serviceName,
-            ImmutableList<Tag> tags,
+            ImmutableList<AwsTag> tags,
             CancellationToken token)
         {
             var taskArns = await ListTaskArns(ecsClient, cluster, serviceName, token);
             var tasks = await DescribeTasks(ecsClient, cluster, taskArns, token);
             // only return tasks with the exact same tags as the filter
-            var tasksWithTags = tasks.Where(task => task.Tags.IsSame(tags, TagComparer)).ToList();
+            var tasksWithTags = tasks.Where(task => task.Tags.Select(t => new AwsTag(t.Key, t.Value)).Diff(tags).Count == 0).ToList();
```
Arkatufus, line 93: This is the actual fix, it was a porting error from Scala.

When this is ported, it is assumed that Scala `list[T].diff()` works by returning a distinct set of items from both the left and right list, so it is assumed that the line `tags.diff(ecsTags).isEmpty` literally means `tags` items have to match `ecsTags`.

Turns out `list[T].diff()` returns items from the left that does not have a match on the right, all it does is a linear walk on the right list and for every item in the right list removes the first matching item on the left.

So the actual correct port for the code `tags.diff(ecsTags).isEmpty` is actually "Is `tags` a subset of `ecsTags`"

Arkatufus (commented):
self-review

PR #2898, opened 2024-11-08
Modernize Coordination.KubernetesApi code

Arkatufus:
## Changes

Modernize Akka.Coordination.KubernetesApi code, there should not be any behavior changes.
* Implement actual async-await code by removing `ContinueWith` and using the `async` keyword
* Remove `ContinueWith` and move the success path into `PipeTo` instead.


Aaronontheweb (approved):
LGTM

PR #3019, opened 2024-12-24
fix `AzureDiscoveryGuardian` child actors created incorrectly

Aaronontheweb:
## Changes

The `AzureDiscoveryGuardian` was creating all of its children as `/user` top-level actors, rather than as actual children under the `/system` hierarchy. This caused them to appear in tracing and for these actors to leak when the discovery plugin restarted or shut itself down.

## Checklist

For significant changes, please ensure that the following have been completed (delete if not relevant):

* [x] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).
* [x] This change follows the [Akka.NET Wire Compatibility Guidelines](https://getakka.net/community/contributing/wire-compatibility.html).
* [x] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).


Aaronontheweb (commented):
These were supposed to be children based on the shutdown code

PR #3285, opened 2025-04-05
Minor fix to contact point seed-nodes log

tstojecki:
## Changes
Fixing bootstrap coordinator's contact point log entry reporting fromAddress instead of seed-nodes


Aaronontheweb (approved):
LGTM

Aaronontheweb:
Nice catch!

Issue #3361, opened 2025-06-23
Cluster bootstrap fails when returned hostname casing differs


Aaronontheweb:
# Cluster bootstrap fails when hostname casing differs  
### Summary
When a node’s **management/remote hostname** differs in letter-case from the value returned by `Akka.Discovery` (Azure Table Storage or DNS), Akka.Management treats them as two distinct addresses.  
This causes the `SelfAwareJoinDecider` to refuse the self-join and the bootstrap coordinator to loop forever.

### Logs (abridged & anonymised)

```text
Located service members … [ResolvedTarget(HOST123, 8558, )], filtered to [HOST123:8558]
…
Self contact point [host123:8558] not found in targets ResolvedTarget(HOST123, 8558, )
Exceeded stable margins without locating seed-nodes …

Aaronontheweb:
Resolved via #3362

PR #3362, opened 2025-06-23
Fix hostname case sensitivity issue in cluster bootstrap self-join

Aaronontheweb:
- Modified SelfAwareJoinDecider.HostMatches to use case-insensitive string comparison
- Fixes issue where hostname casing differences between management/remote hostname and what's returned by Akka.Discovery causes self-join to fail
- Added comprehensive test to verify hostname matching works with different case variations
- Resolves issue #3361

### DIFF
--- a/src/management/Akka.Management/Cluster/Bootstrap/SelfAwareJoinDecider.cs
+++ b/src/management/Akka.Management/Cluster/Bootstrap/SelfAwareJoinDecider.cs
@@ -5,6 +5,7 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
+using System;
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/management/Akka.Management/Cluster/Bootstrap/SelfAwareJoinDecider.cs
```diff
@@ -67,7 +68,8 @@ public bool MatchesSelf(ServiceDiscovery.ResolvedTarget target, (string host, in
         public bool HostMatches(string host, ServiceDiscovery.ResolvedTarget target)
         {
             var cleaned = _hostReplaceRegex.Replace(host, "");
-            return host == target.Host || (target.Address?.ToString() ?? "").Contains(cleaned);
+            return string.Equals(host, target.Host, StringComparison.OrdinalIgnoreCase) ||
```
Aaronontheweb, line 71: Key fix is here - DNS is case-insensitive by design and we need to treat it that way inside the join decider implementations.

Arkatufus (approved):
LGTM

PR #3375, opened 2025-07-14
Make AWS ECS Service discovery mechanism public and non-sealed (#3366)

### LINKED ISSUE #3366: AWS ECS Discovery - Allow discovering "neighboring" nodes from multiple ECS Services
**Is your feature request related to a problem? Please describe.**

Yes.

We have two types of Docker containers that need to be part of a single Akka cluster - but as two different roles:
1. "API" containers: responsible for handling synchronous ASP.NET Web API requests. These containers have the Akka cluster role name as `api`. 
2. "Service" containers: responsible for executing long running or resource intensive operations. The "API" container delegates such operations to the Service container using Akka cluster messages. These containers have the Akka cluster role name as `service`. 

Both these container types are deployed in the same ECS Cluster, but as 2 different ECS Services so that each can be scaled out/in independently. 

AWS ECS Discovery configuration allows only one ECS Service name to be configured. Refer https://github.com/akkadotnet/Akka.Management/tree/dev/src/discovery/aws/Akka.Discovery.AwsApi#using-discovery-together-with-akkamanagement-and-clusterbootstrap-1. 

This will not work for us since our containers are in different ECS Services. 

**Describe the solution you'd like**

- Allow discovering "neighboring" nodes from more than one ECS Service that is part of the same ECS Cluster. ECS Service names could be specified as comma separated values. 
- (OR) Alternatively, kindly open the following class to extension by making it public and not sealed https://github.com/akkadotnet/Akka.Management/blob/701dcda4160e21012a6c2ab9d43c1a4e56adece7/src/discovery/aws/Akka.Discovery.AwsApi/Ecs/EcsServiceDiscovery.cs#L22 This is not possible now since it is marked as an `internal sealed class`. 

**Describe alternatives you've considered**

Considered using the decorator pattern to override the following class and provide a custom implementation that discovers nodes from multiple ECS Services. https://github.com/akkadotnet/Akka.Management/blob/701dcda4160e21012a6c2ab9d43c1a4e56adece7/src/discovery/aws/Akka.Discovery.AwsApi/Ecs/EcsServiceDiscovery.cs#L22 However that is not possible now since the class is marked as an `internal sealed class`. 

**Additional context**

We are migrating to AWS ECS Discovery in place of Lighthouse seed nodes. .

sbenedict-redblack:
Fixes #3366

## Changes

Opens the `EcsServiceDiscovery` class to extension by making it public and non-sealed. This will allow consumers of the library to override it and provide custom implementations. 

For example, out of the box, the library supports discovering "neighboring" nodes only from containers running in the same ECS Service. But as described in #3366, "neighboring" nodes may have to be discovered from multiple ECS Services that are part of an ECS Cluster. With this change, consumers of the library can use the decorator pattern to extend and modify the default behavior of the `EcsServiceDiscovery` class.

## Checklist

For significant changes, please ensure that the following have been completed (delete if not relevant):

* [X] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).
* [X] This change follows the [Akka.NET Wire Compatibility Guidelines](https://getakka.net/community/contributing/wire-compatibility.html).
* [X] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).
* [x] Design discussion issue #3366

### Latest `dev` Benchmarks 

Include data from the [relevant benchmark](https://getakka.net/community/contributing/index.html#improve-performance) prior to this change here.

### This PR's Benchmarks

No impact on performance as a result of this PR.

### DIFF
--- a/src/discovery/aws/Akka.Discovery.AwsApi/Ecs/EcsServiceDiscovery.cs
+++ b/src/discovery/aws/Akka.Discovery.AwsApi/Ecs/EcsServiceDiscovery.cs
@@ -19,7 +19,7 @@ using EcsTask = Amazon.ECS.Model.Task;
 
 namespace Akka.Discovery.AwsApi.Ecs
 {
-    internal sealed class EcsServiceDiscovery : ServiceDiscovery
+    public class EcsServiceDiscovery : ServiceDiscovery
     {
         public static readonly EcsTagComparer TagComparer = new ();
         


Aaronontheweb (approved):
LGTM - trivial change and the reasoning is sound.

PR #3378, opened 2025-08-29
Fix race condition in HostingSpecs.WithAkkaManagement test

Aaronontheweb:
## Summary
- Fixed race condition in `HostingSpecs.WithAkkaManagement` test that was causing intermittent failures with "Address already in use" errors
- The test now uses dynamic port allocation instead of hardcoded port 18558

## Changes
- Modified `HostingSpecs.cs` to use `SocketUtil.TemporaryTcpAddress` for dynamic port allocation
- Added retry logic (up to 3 attempts) to handle race conditions where a port might be taken between detection and binding
- Added specific error handling for `InvalidOperationException` when port binding fails

## Why this fix is needed
The test was failing intermittently in CI/CD pipelines when:
- Multiple test instances ran in parallel
- The OS hadn't released port 18558 quickly enough between test runs
- Other processes happened to be using port 18558

## Test Results
- All 14 HostingSpecs test cases pass consistently
- Ran tests multiple times with no failures
- All 55 tests in Akka.Management.Tests pass

This ensures reliable test execution in parallel test runners and CI/CD pipelines.

### DIFF
--- a/src/management/Akka.Management.Tests/HostingSpecs.cs
+++ b/src/management/Akka.Management.Tests/HostingSpecs.cs
@@ -8,6 +8,8 @@ using System;
 using System.Collections.Generic;
 using System.Net;
 using System.Net.Http;
+using System.Net.Sockets;
+using System.Threading;
 using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Hosting;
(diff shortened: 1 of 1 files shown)


Aaronontheweb (commented):
Trying to eliminate race conditions caused by the static ports in this test specifically

PR #3398, opened 2026-02-25
Fix CreateInNotExistAsync bug

### LINKED ISSUE #3397: SBR `lease-majority` fails to acquire lease due to HTTP 409 `ContainerAlreadyExists`, causing `ReverseDownIndirectlyConnected` to down the entire cluster
## Version Information

- **Akka.NET**: 1.5.58
- **Akka.Coordination.Azure**: (via Akka.Management, matching 1.5.x)
- **.NET Runtime**: 10.0.2
- **Platform**: Azure Container Apps (Kubernetes), Ubuntu 24.04
- **Cluster size**: 5 nodes, all with `core` role
- **Azure Storage SDK**: `Azure.Storage.Blobs` (via Akka.Coordination.Azure dependency)

## Describe the bug

When using the `lease-majority` SBR strategy with `Akka.Coordination.Azure`, the lease acquisition can fail because the Azure Blob Storage container creation call (`PUT /{container-name}?restype=container`) returns HTTP 409 `ContainerAlreadyExists`. This is an expected and benign response when the container was already created by a previous run, but the lease implementation treats it as a failure. This causes SBR to reverse its decision to `ReverseDownIndirectlyConnected`, which downs all indirectly connected + all reachable nodes - effectively destroying the entire cluster.

## SBR Configuration

```hocon
akka.cluster.split-brain-resolver {
  active-strategy = lease-majority
  stable-after = 20s
  down-all-when-unstable = on

  lease-majority {
    lease-implementation = akka.coordination.lease.azure
    lease-name = "<service-name>-akka-sbr"
    acquire-lease-delay-for-minority = 2s
    release-after = 40s
    role = core
  }
}
```

## Steps to Reproduce

1. Run a 5-node Akka.NET cluster using `lease-majority` SBR with `akka.coordination.lease.azure`
2. Ensure the Azure Blob Storage container for lease coordination already exists (i.e., the cluster has been running before)
3. Trigger a network partition where 2 of 5 nodes become mutually unreachable (indirectly connected scenario)
4. Wait for `stable-after` period to elapse so SBR fires
5. Observe: one SBR leader attempts to acquire the lease, the blob container creation returns 409, and the lease acquisition is treated as a failure

### Timeline from production logs

```
T+0s     Two nodes mark each other as UNREACHABLE (mutual/bidirectional)
         This creates an "indirectly connected" scenario

T+20s    SBR stable-after deadline reached. Two competing SBR leaders fire simultaneously
         (one on each side of the partition, as expected for indirectly connected)

T+20s    Leader A: PUT /<lease-container>?restype=container -> HTTP 409 ContainerAlreadyExists
         Leader B: Successfully acquires the blob lease
```

### Leader A (failed to acquire lease)

```
Azure.Core: Error response 409 The specified container already exists. (00.0s)
  Server: Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0
  x-ms-error-code: ContainerAlreadyExists

SBR couldn't acquire lease, reverse decision
  [Akka.Cluster.SBR.AcquireLeaseAndDownIndirectlyConnected]
  to [Akka.Cluster.SBR.ReverseDownIndirectlyConnected]

SBR took decision Akka.Cluster.SBR.ReverseDownIndirectlyConnected
  and is downing [Node-A, Node-B, Node-C, Node-D, Node-E] including myself,
  [2] unreachable of [5] members,
  indirectly connected [Node-D, Node-E]
```

**Result**: Leader A downed **all 5 nodes** including itself.

### Leader B (successfully acquired lease)

```
SBR acquired lease for decision [Akka.Cluster.SBR.AcquireLeaseAndDownIndirectlyConnected]

SBR took decision Akka.Cluster.SBR.AcquireLeaseAndDownIndirectlyConnected
  and is downing [Node-D, Node-E] including myself,
  [1] unreachable of [5] members,
  indirectly connected [Node-D, Node-E]
```

**Result**: Leader B correctly downed only the 2 indirectly connected nodes - but this was overridden by Leader A's decision to down everything.

### Outcome

All 5 nodes were downed. All shard coordinators stopped, all ClusterSingletonManagers terminated, all shard regions shut down. Complete cluster destruction from what should have been a 2-node partition.

## Root Cause Analysis

The bug is in `AzureApiImpl.ContainerClient()` (`src/coordination/azure/Akka.Coordination.Azure/Internal/AzureApiImpl.cs`). The method calls `CreateIfNotExistsAsync()` with **no exception handling**:

```csharp
private async Task<BlobContainerClient> ContainerClient()
{
    var serviceClient = /* ... */;
    var client = serviceClient.GetBlobContainerClient(_settings.ContainerName);

    // Make sure that `CreateIfNotExistsAsync()` only get called once for every AzureApi instance
    if (!_initialized)
    {
        await client.CreateIfNotExistsAsync();  // <-- unhandled 409 thrown here
        _initialized = true;
    }

    return client;
}
```

Despite using `CreateIfNotExistsAsync()` (which should handle 409 internally), the Azure SDK has known bugs where this method still throws `RequestFailedException` with status 409 `ContainerAlreadyExists`. Since there is no `try-catch` around this call, the exception propagates unhandled.

### Failure propagation path

1. `ContainerClient()` throws `RequestFailedException` (409)
2. This propagates through whichever method called it (`LeaseResourceExists()`, `GetLeaseResource()`, or `CreateLeaseResource()`)
3. Which propagates through `ReadOrCreateLeaseResource()`
4. In `LeaseActor.cs`, the `PipeTo` failure handler converts it to `Status.Failure`:
   ```csharp
   // LeaseActor.cs - Idle state, ReadRequired case:
   _client.ReadOrCreateLeaseResource(leaseName)
       .ContinueWith(task => new ReadResponse(task.Result))
       .PipeTo(Self, failure: FlattenAggregateException);
   ```
5. The `Status.Failure` is caught by `WhenUnhandled` (since the `PendingRead` state only handles `ReadResponse`, not `Status.Failure`):
   ```csharp
   // LeaseActor.cs - WhenUnhandled:
   case Status.Failure f when @event.StateData is IReplyRequired replyRequired:
       _log.Warning(f.Cause, "Failure communicating with the API server...");
       replyRequired.ReplyTo.Tell(new Status.Failure(f.Cause));
       return GoTo(Idle.Instance).Using(ReadRequired.Instance);
   ```
6. The SBR `LeaseMajority` implementation receives the `Status.Failure`, treats it as a lease acquisition failure, and reverses the decision:
   ```
   SBR couldn't acquire lease, reverse decision
     [AcquireLeaseAndDownIndirectlyConnected] to [ReverseDownIndirectlyConnected]
   ```

### Why `ReverseDownIndirectlyConnected` destroys the cluster

The `ReverseDownIndirectlyConnected` decision (from `DowningStrategy.cs`) downs indirectly connected nodes **plus all reachable nodes**:

```csharp
// ReverseDownIndirectlyConnected downs indirectly connected + all reachable nodes
return downable.Intersect(IndirectlyConnected)
    .Union(downable.Except(Unreachable));
```

In a 5-node cluster with 2 indirectly connected and 3 reachable nodes, this downs all 5 nodes.

## Expected Behavior

The blob container creation returning HTTP 409 `ContainerAlreadyExists` should be handled as an expected/benign response. The lease acquisition should proceed to the actual blob lease acquisition step rather than failing.

This would allow the `lease-majority` strategy to work correctly: only one leader acquires the lease, that leader makes the correct downing decision, and the other leader (failing to acquire the lease) correctly reverses and downs its own side.

## Suggested Fix

Wrap the `CreateIfNotExistsAsync()` call in `AzureApiImpl.ContainerClient()` with a `try-catch` for 409 responses. The code already uses `CreateIfNotExistsAsync()` correctly, but needs to defend against the Azure SDK's known bugs where this method still throws.

### Recommended change to `AzureApiImpl.ContainerClient()`

```csharp
private async Task<BlobContainerClient> ContainerClient()
{
    var serviceClient = _settings.AzureCredential != null && _settings.ServiceEndpoint != null
        ? new BlobServiceClient(_settings.ServiceEndpoint, _settings.AzureCredential, _settings.BlobClientOptions)
        : new BlobServiceClient(_settings.ConnectionString);

    var client = serviceClient.GetBlobContainerClient(_settings.ContainerName);

    if (!_initialized)
    {
        try
        {
            await client.CreateIfNotExistsAsync();
        }
        catch (RequestFailedException ex) when (ex.Status == 409)
        {
            // Safe to ignore - container already exists.
            // CreateIfNotExistsAsync should handle this internally but has known SDK bugs
            // where it still throws on ContainerAlreadyExists or ContainerBeingDeleted.
            // See: https://github.com/Azure/azure-sdk-for-net/issues/28549
        }
        _initialized = true;
    }

    return client;
}
```

This is a minimal, targeted fix — only the unprotected `CreateIfNotExistsAsync()` call needs the guard. The Azure SDK issues ([#28549](https://github.com/Azure/azure-sdk-for-net/issues/28549), [#18000](https://github.com/Azure/azure-sdk-for-net/issues/18000), [#39845](https://github.com/Azure/azure-sdk-for-net/issues/39845), [#9758](https://github.com/Azure/azure-sdk-for-net/issues/9758)) document why `CreateIfNotExistsAsync` alone is insufficient.

### Benign 409 error codes to handle

The following Azure Blob Storage 409 error codes are benign in the context of container creation for lease coordination:

| Error Code | Description | Why it's benign |
|---|---|---|
| `ContainerAlreadyExists` | Container was already created by a previous run or another node | Normal steady-state; container is ready to use |
| `ContainerBeingDeleted` | Container with same name was recently deleted and Azure hasn't finished cleanup (~30s) | Transient; retry after a short delay and the container will be available |

### Transient errors that should be retried (not treated as permanent failures)

| HTTP Status | Error Code | Description |
|---|---|---|
| 429 | `ServerBusy` | Azure Storage rate limiting / throttling |
| 503 | `ServerBusy` | Azure Storage temporarily overloaded |
| 500 | `InternalError` | Transient Azure-side failure |

These transient errors should trigger a retry with backoff rather than causing the lease acquisition to fail permanently. A failed lease acquisition in the SBR context has catastrophic consequences (`ReverseDownIndirectlyConnected` downing the entire cluster), so the retry policy should be generous.

## Workaround

Users can work around this by switching to `keep-majority` SBR strategy, which does not depend on an external lease and avoids this code path entirely:

```hocon
akka.cluster.split-brain-resolver {
  active-strategy = keep-majority
  stable-after = 20s
}
```

## Impact

- **Severity**: Critical - can cause complete cluster destruction during routine network partitions
- **Trigger**: Any indirectly connected partition in a cluster that has been running before (i.e., the blob container already exists)
- **Scope**: All users of `lease-majority` with `akka.coordination.lease.azure`

### Azure SDK for .NET — `CreateIfNotExistsAsync` reliability bugs

The Azure Storage Blobs SDK has several known issues where `BlobContainerClient.CreateIfNotExistsAsync()` does not properly suppress 409 responses, which is directly relevant to the recommended fix:

- [Azure/azure-sdk-for-net#28549](https://github.com/Azure/azure-sdk-for-net/issues/28549) — `CreateIfNotExistsAsync` throws `RequestFailedException` (409) instead of returning silently
- [Azure/azure-sdk-for-net#18000](https://github.com/Azure/azure-sdk-for-net/issues/18000) — `CreateIfNotExistsAsync` intermittently fails with 409 `ContainerAlreadyExists`
- [Azure/azure-sdk-for-net#39845](https://github.com/Azure/azure-sdk-for-net/issues/39845) — `NullReferenceException` from `CreateIfNotExistsAsync` under certain conditions
- [Azure/azure-sdk-for-net#9758](https://github.com/Azure/azure-sdk-for-net/issues/9758) — Additional `CreateIfNotExistsAsync` reliability issues

These bugs reinforce why the fix should use both `CreateIfNotExistsAsync` **and** an explicit `catch (RequestFailedException ex) when (ex.Status == 409)` guard.

Arkatufus:
Fixes #3397

### Problem

`AzureApiImpl.ContainerClient()` called `CreateIfNotExistsAsync()` with no exception handling. The Azure SDK has [known bugs](https://github.com/Azure/azure-sdk-for-net/issues/28549) where this method still throws `RequestFailedException(409)` for `ContainerAlreadyExists`. The unhandled exception propagates through the lease actor to the Split Brain Resolver, which reverses its decision to `ReverseDownIndirectlyConnected` and downs the entire cluster.

**Propagation path:** `ContainerClient()` throws 409 → `LeaseResourceExists()` catch block misidentifies it as a blob-level error → re-thrown as `LeaseException` → `ReadOrCreateLeaseResource()` (no catch) → `LeaseActor` PipeTo → `Status.Failure` → SBR receives failure → `ReverseDownIndirectlyConnected` → all nodes downed.

### Fix

**Replaced `CreateIfNotExistsAsync()` with `CreateAsync()` + explicit 409 handling** in `ContainerClient()`:
- `ContainerAlreadyExists` (benign) — silently handled, set `_initialized = true`
- All other errors — wrapped in a new internal `ContainerInitializationException`

**New `ContainerInitializationException`** — an internal exception type that distinguishes container-level errors from blob-level `RequestFailedException`. This prevents the existing `catch (RequestFailedException)` blocks (designed for blob operations) from mishandling container creation errors.

**Catch blocks added to `LeaseResourceExists()`, `CreateLeaseResource()`, and `GetLeaseResource()`** — each catches `ContainerInitializationException` and returns its natural "try again" value (`false`, `null`, `null`), which feeds back into the existing retry loop in `ReadOrCreateLeaseResource()`.

### Files Changed

| File | Change |
|------|--------|
| `src/coordination/azure/Akka.Coordination.Azure/Internal/AzureApiImpl.cs` | Add `ContainerInitializationException`; rewrite `ContainerClient()`; add catch blocks to caller methods |
| `src/coordination/azure/Akka.Coordination.Azure.Tests/AzureApiSpec.cs` | Add regression tests for container-already-exists scenario |

### Test Plan

- [x] `ShouldHandleContainerAlreadyExists` — reproduces the exact #3397 scenario: second `AzureApiImpl` instance hits pre-existing container, verifies it succeeds instead of throwing
- [x] `MultipleInstancesShouldAcquireDifferentLeases` — simulates multi-node cluster where two instances race to create the same container
- [x] All 47 existing tests pass (1 pre-existing skip), 0 regressions

### DIFF
--- a/src/coordination/azure/Akka.Coordination.Azure.Tests/AzureApiSpec.cs
+++ b/src/coordination/azure/Akka.Coordination.Azure.Tests/AzureApiSpec.cs
@@ -104,5 +104,55 @@ namespace Akka.Coordination.Azure.Tests
             var response = await _underTest.RemoveLease(LeaseName);
             response.Should().Be(Done.Instance);
         }
+
+        // Regression test for https://github.com/akkadotnet/Akka.Management/issues/3397
+        // A second AzureApiImpl instance (_initialized = false) encountering a container that
+        // already exists must handle the 409 ContainerAlreadyExists gracefully instead of
+        // throwing a LeaseException that propagates up to the Split Brain Resolver.
+        [Fact(DisplayName = "Should handle ContainerAlreadyExists when a second instance starts")]
+        public async Task ShouldHandleContainerAlreadyExists()
+        {
+            // First instance creates the container and lease blob
+            var firstLease = await _underTest.ReadOrCreateLeaseResource(LeaseName);
+            firstLease.Owner.Should().BeNull();
+
+            // Second instance has _initialized = false, so ContainerClient() will call
+            // CreateAsync() and receive a 409 ContainerAlreadyExists from Azure.
+            // Before the fix, this threw LeaseException and crashed the lease actor.
+            var secondInstance = new AzureApiImpl(Sys, _settings);
+            var secondLease = await secondInstance.ReadOrCreateLeaseResource(LeaseName);
+            secondLease.Owner.Should().BeNull();
+            secondLease.Version.Should().NotBeNull();
+        }
+
+        // Verifies that multiple independent AzureApiImpl instances can operate concurrently
+        // against the same container — the typical scenario in a multi-node Akka.NET cluster
+        // where each node creates its own AzureApiImpl.
+        [Fact(DisplayName = "Multiple instances should acquire different leases against same container")]
+        public async Task MultipleInstancesShouldAcquireDifferentLeases()
+        {
+            const string leaseName1 = "lease-multi-1";
+            const string leaseName2 = "lease-multi-2";
+            const string owner1 = "node-1";
+            const string owner2 = "node-2";
+
+            var instance1 = new AzureApiImpl(Sys, _settings);
+            var instance2 = new AzureApiImpl(Sys, _settings);
+
+            // Both instances create their leases (both will try CreateAsync on the container)
+            var lease1 = await instance1.ReadOrCreateLeaseResource(leaseName1);
+            var lease2 = await instance2.ReadOrCreateLeaseResource(leaseName2);
+
+            // Both should succeed — one creates the container, the other gets 409 and handles it
+            lease1.Owner.Should().BeNull();
+            lease2.Owner.Should().BeNull();
+
+            // Both instances should be able to update their respective leases
+            var update1 = await instance1.UpdateLeaseResource(leaseName1, owner1, lease1.Version, DateTimeOffset.UtcNow);
+            update1.Should().BeOfType<Right<LeaseResource, LeaseResource>>();
+
+            var update2 = await instance2.UpdateLeaseResource(leaseName2, owner2, lease2.Version, DateTimeOffset.UtcNow);
+            update2.Should().BeOfType<Right<LeaseResource, LeaseResource>>();
+        }
     }
 }
--- a/src/coordination/azure/Akka.Coordination.Azure/Internal/AzureApiImpl.cs
+++ b/src/coordination/azure/Akka.Coordination.Azure/Internal/AzureApiImpl.cs
@@ -21,6 +21,24 @@ using Newtonsoft.Json;
 
 namespace Akka.Coordination.Azure.Internal
 {
+    /// <summary>
+    /// Thrown when Azure Blob container initialization fails.
+    /// Used to distinguish container-level errors from blob-level <see cref="RequestFailedException"/>
+    /// so that callers can retry appropriately.
+    /// </summary>
+    internal sealed class ContainerInitializationException : Exception
+    {
+        public int StatusCode { get; }
+        public string? ErrorCode { get; }
+
+        public ContainerInitializationException(string message, int statusCode, string? errorCode, Exception innerException)
+            : base(message, innerException)
+        {
+            StatusCode = statusCode;
+            ErrorCode = errorCode;
+        }
+    }
+
     internal sealed class AzureApiImpl: IAzureApi
     {
         private readonly AzureLeaseSettings _settings;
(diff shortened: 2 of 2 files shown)


Aaronontheweb (approved):
LGTM

PR #3406, opened 2026-03-09
Add failing tests proving premature LeaseAcquired bug (#3402)

### LINKED ISSUE #3402: Premature `LeaseAcquired` in `LeaseActor` Granting state causes shard-level split brain
`LeaseActor` sends `LeaseAcquired` to the caller before a retry write completes in the `Granting` state. This allows a shard to start running without a valid lease, without heartbeats, and without lease-loss detection — enabling two nodes to run the same shard simultaneously.

## Affected versions

All versions of `Akka.Coordination.Azure` (bug predates 1.5.61).

## Bug location

`LeaseActor.cs` — `Granting` state handler, conflict response path when the blob has no owner.

**Before fix:**
```csharp
// Granting state, Left (conflict) response, owner is null:
if (leftResponse.Owner is null)
{
    // ...version check...
    who.Tell(LeaseAcquired.Instance);  // <-- BUG: sent before retry completes
    _client.UpdateLeaseResource(leaseName, _ownerName, version)
        .ContinueWith(t => new WriteResponse(t.Result))
        .PipeTo(Self, failure: FlattenAggregateException);
    return Stay();
}
```

## How it happens

1. Node B writes to acquire the lease → gets a **write conflict** (`Left` response)
2. The conflict read shows the blob has **no owner** (another node wrote and released between B's read and write)
3. `LeaseActor` sends `LeaseAcquired` to the caller **immediately** — the `Ask` completes and the shard starts
4. `LeaseActor` retries the write but stays in `Granting` (not `Granted`)

If the retry also conflicts and another node now owns the lease:

5. `LeaseActor` sends `LeaseTaken` to `who` — but the temp `Ask` actor already completed and stopped, so it goes to **dead letters**
6. FSM transitions to `Idle` with `ReadRequired`

## Impact

After the premature `LeaseAcquired`, the shard is running but:

| What should happen | What actually happens |
|---|---|
| `localGranted = true` | `localGranted` stays `false` (only set on `Right` response, line 349) |
| Heartbeat timer starts | Timer never starts (only starts on transition **to** `Granted`) |
| `LeaseLost` callback is preserved | Callback is lost — `OperationInProgress` data (which holds the callback) is replaced with `ReadRequired` when FSM goes to `Idle` |
| Lease loss is detected | No detection — `LeaseActor` is idle, no heartbeats, no callback |

### Shard-level split brain

Since no heartbeats are being written, the blob's timestamp freezes. After the TTL expires (default 96 seconds), **another node can acquire the same shard's lease**. Both nodes now run the same shard simultaneously.

When the shard eventually stops (HandOff or shutdown), `ReleaseLeaseIfNeeded()` sends `Release` to the `LeaseActor`, which is in `Idle` state. The `WhenUnhandled` handler returns `InvalidReleaseRequest` — the **blob is never cleared**. Other nodes must wait for TTL even after this node shuts down.

### Note on `CheckLease()`

`Akka.Cluster.Sharding` does **not** poll `CheckLease()`. It uses a callback-based pattern:
```csharp
// Shard.cs
var result = await lease.Acquire(reason => { self.Tell(new LeaseLost(reason)); });
```

The `LeaseLost` callback is the only mechanism for the shard to detect lease loss. Since the callback is lost when the FSM drops to `Idle`, the shard has **no way** to know the lease was never actually acquired.

## Trigger conditions

This requires a specific but plausible sequence:

1. Write conflict in `Granting` state (ETag mismatch — another node wrote to the same blob)
2. The conflict read shows no owner (the other node released between our read and write)
3. The retry write also fails (another node takes the lease before retry completes)

This is more likely during **rolling upgrades** with many shards rebalancing simultaneously, where multiple nodes race to acquire leases.

Aaronontheweb:
## Summary

Adds deterministic reproduction tests for the premature `LeaseAcquired` split-brain bug in both Azure and Kubernetes `LeaseActor` implementations.

**These tests intentionally FAIL** to prove the bug exists. The fix will come in a subsequent commit.

### The Bug

When a CAS conflict occurs during the `Granting` state and the blob/configmap has no owner (version moved on), `LeaseActor` sends `LeaseAcquired` to the caller **before** the retry write completes ([K8s line 365](https://github.com/akkadotnet/Akka.Management/blob/dev/src/coordination/kubernetes/Akka.Coordination.KubernetesApi/LeaseActor.cs#L365), [Azure line 362](https://github.com/akkadotnet/Akka.Management/blob/dev/src/coordination/azure/Akka.Coordination.Azure/LeaseActor.cs#L362)). If the retry then fails (another node takes the lease), the caller already believes it holds the lease — but it doesn't. `localGranted` is never set to `true`, heartbeat never starts.

### Test Results

| Test | K8s | Azure | Expected |
|------|-----|-------|----------|
| `ShouldNotSendPrematureLeaseAcquiredWhenConflictRetryIsStolen` | FAIL | FAIL | Will pass after fix |
| `ShouldGrantLeaseOnlyAfterConflictRetrySucceeds` | PASS | PASS | Happy-path verification |

### Failure Output
```
Expected a message of type LeaseTaken, but received LeaseAcquired
```

This directly proves the premature `LeaseAcquired` is sent before the retry write, confirming the split-brain scenario described in #3402.

### All existing tests pass
- K8s: 47 passed, 1 skipped, 0 failed
- Azure: 45 passed, 1 skipped, 0 failed

Fixes #3402

Closes #3403

### DIFF
--- a/src/coordination/azure/Akka.Coordination.Azure.Tests/LeaseActorSpec.cs
+++ b/src/coordination/azure/Akka.Coordination.Azure.Tests/LeaseActorSpec.cs
@@ -393,6 +393,86 @@ namespace Akka.Coordination.Azure.Tests
             // TODO this could accumulate senders and reply to all, atm it'll log saying previous action hasn't finished
         }
 
+        // Regression test: when a write conflict occurs in Granting state and the blob has no owner,
+        // LeaseAcquired must NOT be sent until the retry write succeeds. Sending it prematurely causes
+        // the shard to start without localGranted=true and without heartbeats, leading to a shard-level
+        // split brain if another node takes the lease before the retry completes.
+        [Fact(DisplayName = "LeaseActor should not send LeaseAcquired prematurely on conflict retry in Granting state")]
+        public void ShouldNotSendLeaseAcquiredPrematurelyOnConflictRetry()
+        {
+            RunTest(() =>
+            {
+                // Start acquiring: read returns empty lease
+                UnderTest.Tell(new LeaseActor.Acquire(), Sender);
+                LeaseProbe.ExpectMsg(LeaseName);
+                LeaseProbe.Reply(new LeaseResource("", CurrentVersion, CurrentTime));
+
+                // First write attempt
+                UpdateProbe.ExpectMsg((OwnerName, CurrentVersion));
+
+                // Conflict: version moved but no owner (another node wrote and released)
+                var conflictVersion = new ETag((CurrentVersionCount + 3).ToString());
+                UpdateProbe.Reply(
+                    new Left<LeaseResource, LeaseResource>(
+                        new LeaseResource("", conflictVersion, CurrentTime)));
+
+                // LeaseAcquired must NOT have been sent yet — the retry hasn't completed
+                SenderProbe.ExpectNoMsg(TimeSpan.FromMilliseconds(100));
+
+                // Retry write is issued with the new version
+                UpdateProbe.ExpectMsg((OwnerName, conflictVersion));
+
+                // Retry also conflicts, but this time another node owns it
+                var stolenVersion = new ETag((CurrentVersionCount + 5).ToString());
+                UpdateProbe.Reply(
+                    new Left<LeaseResource, LeaseResource>(
+                        new LeaseResource("another-node", stolenVersion, CurrentTime)));
+
+                // Now the caller should get LeaseTaken (not a stale LeaseAcquired)
+                SenderProbe.ExpectMsg<LeaseActor.LeaseTaken>();
+                Granted.Value.Should().BeFalse();
+            });
+        }
+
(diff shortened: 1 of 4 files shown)


Aaronontheweb:
Integrated Gregorius' original fix commits into this branch and extended them to Kubernetes for parity.\n\nIncluded with attribution via cherry-pick:\n- 611112cd4474c8a5910b98586d6f8d5b6c77f69d (`Fix premature LeaseAcquired in LeaseActor Granting state`)\n- a2ed0e2337c6440393d9936cd3fe02ba7c1a5f91 (`Harden test`)\n\nAdditional commit in this branch:\n- Kubernetes equivalent fix for premature `LeaseAcquired` in the null-owner conflict retry path.\n\nThis keeps the repro-first record intact while preserving prior work and author credit.

PR #3407, opened 2026-03-09
Add failing tests proving heartbeat surrender bug (#3404)

### LINKED ISSUE #3404: Transient heartbeat failure immediately surrenders lease, causing unnecessary ~96s shard outage
A single transient failure during a heartbeat write (network timeout, Azure 500/503, DNS blip) causes the `LeaseActor` to immediately surrender a lease it still legitimately holds. The blob is left with a stale owner, blocking all other nodes from acquiring the shard for up to 96 seconds.

## Affected versions

All versions of `Akka.Coordination.Azure`. The code contains an existing `// FIXME` comment acknowledging this issue.

## Bug location

[`LeaseActor.cs`](src/coordination/azure/Akka.Coordination.Azure/LeaseActor.cs) — `Granted` state handler, `Status.Failure` case (lines 407–412):

```csharp
case Status.Failure failure:
    // FIXME, retry if timeout far enough off: https://github.com/lightbend/akka-commercial-addons/issues/501
    _log.Warning(failure.Cause, "Failure during heartbeat to lease. Lease assumed to be released.");
    localGranted.GetAndSet(false);
    ExecuteLeaseLockCallback(leaseLost, failure.Cause);
    return GoTo(Idle.Instance).Using(ReadRequired.Instance);
```

## How it happens

1. Node A holds the lease for shard 13 and is in the `Granted` state, heartbeating every 12 seconds
2. A heartbeat write times out or fails transiently (Azure 503, network blip, DNS timeout)
3. `UpdateLeaseResource` throws `LeaseTimeoutException` → PipeTo wraps it as `Status.Failure`
4. The `Granted` handler receives `Status.Failure` and **immediately surrenders**:
   - `localGranted = false`
   - `LeaseLost` callback fires → Shard stops
   - FSM goes to `Idle` → heartbeats stop

**The blob was never modified.** It still shows Node A as owner with the last successful heartbeat timestamp.

## Why this is wrong

`Status.Failure` means the write **did not complete** — the blob is unchanged. This is fundamentally different from the `Left` (conflict) case where Azure confirmed another node owns the blob.

The `LeaseTimeoutException` message itself acknowledges this: *"It is not known if the update happened."* If the write did silently succeed, the blob has a fresh timestamp and retrying is harmless. If it didn't, retrying is necessary. In neither case is immediate surrender correct.

## Impact

```
1. Heartbeat write times out          → Status.Failure
2. Node A surrenders lease            → localGranted=false, LeaseLost callback, Idle
3. Shard stops on Node A              → no heartbeats being written
4. Blob still shows Node A as owner   → timestamp frozen at last successful heartbeat
5. Node B tries to acquire shard 13   → reads blob → Owner=A, not timed out → LeaseTaken
6. Node B retries every 5s            → blocked until TTL expires
7. After 96s TTL                      → Node B can finally take over
```

A single transient timeout causes an unnecessary **~96-second shard outage**. During rolling deployments this compounds across multiple shards.

## Contrast with the `Left` (conflict) case

The heartbeat conflict handler (lines 401–405) is **correct**:

```csharp
case WriteResponse {Response: Left<LeaseResource, LeaseResource> resource}:
    _log.Warning("Conflict during heartbeat to lease {0}. Lease assumed to be released.", resource.Value);
    localGranted.GetAndSet(false);
    ExecuteLeaseLockCallback(leaseLost, null);
    return GoTo(Idle.Instance).Using(ReadRequired.Instance);
```

A `Left` response means the write got a 412/409 **and** Azure was reachable — `UpdateLeaseResource` performed a fresh GET and the blob shows a different owner. Surrendering is the right thing to do. No change needed here.

Aaronontheweb:
## Summary
- Adds deterministic reproduction tests for Bug #3404 in both Azure and Kubernetes `LeaseActor` specs.
- The tests prove a transient heartbeat `Status.Failure` currently causes immediate lease surrender instead of retrying within the TTL safety window.
- These tests intentionally fail to establish proof before any fix is applied.

## What these tests assert
- After successful acquisition and heartbeat start, a single transient heartbeat failure should **not**:
  - set `Granted` to `false`,
  - invoke `leaseLostCallback`, or
  - stop heartbeat retry attempts.
- Instead, actor should remain granted and schedule another heartbeat write attempt.

## Current behavior (bug)
Both implementations immediately surrender lease on `Status.Failure` while in `Granted`:
- Azure: `LeaseActor.cs` in `Granted` handler (`Status.Failure` branch)
- Kubernetes: `LeaseActor.cs` in `Granted` handler (`Status.Failure` branch)

Observed failure in both test suites:
- Expected retry update message within heartbeat interval window
- Actual: no retry; actor transitions to idle and marks lease lost

## Why this PR
Previous fix attempts went straight to implementation. This PR is repro-only to build a deterministic, reviewable proof that the bug is real in both providers before proposing fixes.

Fixes #3404

Closes #3405

### DIFF
--- a/src/coordination/azure/Akka.Coordination.Azure.Tests/LeaseActorSpec.cs
+++ b/src/coordination/azure/Akka.Coordination.Azure.Tests/LeaseActorSpec.cs
@@ -363,6 +363,29 @@ namespace Akka.Coordination.Azure.Tests
             });
         }
 
+        [Fact(DisplayName = "Bug #3404: should retry heartbeat on transient failure without surrendering lease")]
+        public void ShouldRetryHeartbeatOnTransientFailureWithoutSurrenderingLease()
+        {
+            RunTest(() =>
+            {
+                Exception? callbackError = null;
+                AcquireLease(e => callbackError = e);
+                ExpectHeartBeat();
+                Granted.Value.Should().BeTrue();
+
+                UpdateProbe.ExpectMsg((OwnerName, CurrentVersion));
+                var transientFailure = new LeaseException("Transient failure");
+                UpdateProbe.Reply(new Status.Failure(transientFailure));
+
+                // Correct behavior: retain lease and retry while TTL still allows it.
+                Granted.Value.Should().BeTrue();
+                callbackError.Should().BeNull();
+
+                var retry = UpdateProbe.ExpectMsg<(string, ETag)>(LeaseSettings.TimeoutSettings.HeartbeatInterval * 3);
+                retry.Should().Be((OwnerName, CurrentVersion));
+            });
+        }
+
         [Fact(DisplayName = "lock should be acquire-able after heart beat conflict")]
         public void LockShouldAcquireAfterHeartBeatConflict()
         {
--- a/src/coordination/azure/Akka.Coordination.Azure.Tests/LeaseActorSpec.cs
+++ b/src/coordination/azure/Akka.Coordination.Azure.Tests/LeaseActorSpec.cs
@@ -324,18 +324,13 @@ namespace Akka.Coordination.Azure.Tests
         {
             RunTest(() =>
             {
-                var failure = new LeaseException("Failed to communicate with API server");
                 AcquireLease();
                 ExpectHeartBeat();
                 Granted.Value.Should().BeTrue();
 
-                UpdateProbe.ExpectMsg((OwnerName, CurrentVersion));
-                IncrementVersion();
-                UpdateProbe.Reply(new Status.Failure(failure));
-                AwaitAssert(() =>
-                {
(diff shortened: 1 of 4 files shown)


Aaronontheweb:
Integrated Gregorius' original Azure heartbeat-retry fix into this branch and extended it to Kubernetes for parity.\n\nIncluded with attribution via cherry-pick:\n- 0cf2b4e3ea9be5a56686d2e302b2875bc8af4a2c (`Retry heartbeat on transient failure instead of immediately surrendering lease`)\n\nAdditional commit in this branch:\n- Kubernetes heartbeat retry implementation + matching test updates.\n\nThis preserves the repro-first evidence trail while carrying forward prior work with proper credit.

PR #3418, opened 2026-03-18
Fix heartbeat self-conflict causes unnecessary lease release after transient timeout

Arkatufus:
### Summary

- When a heartbeat PUT succeeds on the API server but times out on the client, the retry uses a stale version and gets a CAS conflict. If the conflict response shows the same owner, the lease was never actually lost — but the actor was unconditionally releasing it.
- The `Left<>` handler in the `Granted` state now checks whether the conflict owner matches the current node. Same owner: update the version and continue heartbeating. Different owner: release the lease (unchanged behavior).
- Affects both Kubernetes and Azure lease actor implementations.

### Changes

**`LeaseActor.cs` (both Kubernetes and Azure):** Added an owner equality check before the existing unconditional release in the heartbeat conflict handler. Uses `string.Equals` with `StringComparison.Ordinal` for strict matching.

**`LeaseActorSpec.cs` (both Kubernetes and Azure):** Added 3 reproduction tests per implementation:
- `HeartbeatSelfConflictAfterTimeoutShouldStayGranted` — timeout → retry → self-conflict → stays granted
- `HeartbeatSelfConflictShouldNotCallLeaseLostCallback` — lease-lost callback must not fire on self-conflict
- `HeartbeatConflictWithDifferentOwnerAfterTimeoutShouldRelease` — guards that genuine conflicts still release

### Test plan

- [x] All 6 new tests pass (3 Kubernetes + 3 Azure)
- [x] All existing lease actor tests pass on both platforms
- [x] Azure integration tests pass with Docker/Azurite


src/coordination/kubernetes/Akka.Coordination.KubernetesApi/LeaseActor.cs
```diff
@@ -398,6 +398,18 @@ public LeaseActor(IKubernetesApi client, LeaseSettings settings, string leaseNam
                         return Stay().Using(gv.Copy(version: resource.Value.Version, lastSuccessfulHeartbeat: DateTime.UtcNow));
                     
                     case WriteResponse {Response: Left<LeaseResource, LeaseResource> resource}:
+                        if (string.Equals(resource.Value.Owner, _ownerName, StringComparison.Ordinal))
+                        {
+                            // Self-conflict: our previous PUT succeeded on the server but timed out
+                            // on the client, so we retried with a stale version. The lease is still ours —
+                            // update the version and continue heartbeating.
+                            _log.Warning(
```
Aaronontheweb, line 406: `resource.Value.Owner` and `_ownerName` are always the same so we could clean this log message up and simplify it.

Aaronontheweb (changes requested):
Minor nitpick

Aaronontheweb (approved):
LGTM

PR #3444, opened 2026-08-17
Add Akka.Discovery.Redis service discovery plugin

Aaronontheweb:
## Summary

Brings the Redis-based service discovery plugin into Akka.Management as a first-class, supported package, alongside the existing `Akka.Discovery.Azure`, `Akka.Discovery.KubernetesApi`, and friends. It originated in a personal .NET Aspire plugin repo and had not been battle-tested outside local Aspire development, so this PR also hardens it against the mature `Akka.Discovery.Azure` sibling.

## What's here

New `Akka.Discovery.Redis` package under `src/discovery/redis/` (`netstandard2.0;net6.0`). Each node registers itself under a Redis key with a TTL and refreshes it on a heartbeat; peers discover each other by scanning those keys. Dead nodes are removed by Redis key expiry, so there is no `PruneActor` — Redis does the physical removal natively (Table Storage can't, which is why `Akka.Discovery.Azure` needs one).

## Hardening (vs. Akka.Discovery.Azure)

- **Lazy connection ownership** — the `ConnectionMultiplexer` is created inside the guardian actor, not in the discovery constructor. An unreachable Redis at startup no longer blocks or fails ActorSystem startup; the guardian retries with backoff and recovers when Redis comes back.
- **Backoff + per-operation timeout** on every Redis operation, replacing the previous zero-delay hot-retry loop. New HOCON knobs: `operation-timeout`, `retry-backoff`, `max-retry-backoff`.
- **Stale-node filtering** — lookups exclude entries older than `stale-ttl-threshold` (defaults to `min(3×heartbeat, ttl)`), so a just-died node drops out of discovery results faster than the physical key TTL.
- Fixed the documented `public-hostname` → `akka.remote.dot-netty.tcp.public-hostname` fallback (it previously jumped straight to `Dns.GetHostName()`).
- Added read-only mode and `IDiscoveryOptions` support for first-class Akka.Hosting integration.

## Tests

38 tests (25 unit + 13 Testcontainers.Redis integration, tagged `Category=Integration`, skip gracefully when Docker is absent):

- settings parsing/validation/`public-hostname` fallback, hosting-extension config, discovery-method load guard, serialization compatibility (missing/unknown fields)
- real-Redis client ops, guardian actor behavior, **multi-node convergence** (3 nodes each discover all peers), TTL expiry, and **resilience** (startup survives an unreachable Redis, then recovers once Redis is reachable)


src/discovery/redis/Akka.Discovery.Redis/Properties/FriendsOf.cs
```diff
@@ -0,0 +1,9 @@
+// -----------------------------------------------------------------------
+//  <copyright file="FriendsOf.cs" company="Akka.NET Project">
+//      Copyright (C) 2013-2026 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//  </copyright>
+// -----------------------------------------------------------------------
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Akka.Discovery.Redis.Tests")]
```
Aaronontheweb, line 9: you can declare `Friends` inside MSBuild / `.csproj` metadata now

